PR #6, opened 2015-01-05
Added romanian translation for all values

tudorbarascu:


3nids:
Hi Tudor,

Thanks for this!

Just a small tip: can you please rebase your commits before pulling a request?
This would avoid having too many commits (especially if all are about the same -- i.e. translation).
You can do ```git rebase -i abc1234^``` where abc1234 is your first commit you want to merge.
Then, you can replace all but the first "pick" by a "s" for squash. This will squash all the commits into one.

Best wishes

tudorbarascu:
Hi Denis and a happy new year.
Will be happy to do rebasing from now on. Thanks for the tip!
Best wishes.

PR #9, opened 2015-01-15
Possible combinations for pipe diameters of PE materials.

tudorbarascu:
Thee combinations are based on E.U. standards that I could find.
I haven't found a final way of including them in qwat logic yet but it's a stepping stone and it should help us later on.
All the best


3nids:
Thanks!
But how would you use this table?
Wouldn't it be better to integrate this with vl_pipe_material?

For the valves, it might be better to do a separate PR. Also, "vanne réseau" sounds better than "vanne de réseau".

tudorbarascu:
Yes, it would be better to integrate this with vl_pipe_material, but for now, as there are so many combinations I didn't wan't to mess with the current workflow when choosing materials.
I wanted to give you a look at the table as you have different perspective. I'll repull with the "vanne reseau" as you're right, it sounds better.
All the best

tudorbarascu:
Should vanne de branchement --> vanne branchement?

PR #12, opened 2015-01-22
completing PE values, vl_valve function

tudorbarascu:
Everything should be alright now. It would be nice if we could erase values 5502 and 5505, PE of 100 and 150 as they shouldn't exist as they're not on any standard. But if they're there then you must be using them. Do tell if I can remove them,

All the best.


3nids:
indeed, I have no instances of PE 100 and 150, I will take care of removing them.
thanks for your work!

tudorbarascu:
Ok, great, In the near future, a next step that I will work on would be creating some triggers to automatically set the id_material (for PE) when a user saves a pipe. (e.g. saving PE pipe with 100 diameter and a certain nominal pressure should automatically change to the correct id_material. for now, the user has multiple PE 100, but only one fits the bill).
All the best

PR #16, opened 2015-02-04
enable audit for vl_pressurecontrol_type_ + fix displayname_ function to take into account materials without known diameter

tudorbarascu:


tudorbarascu:
Do you think we should also have a value for PE80 material with unknown diameter? We do have for other types of materials and for PE (PE100).
We could also leave it as PE generic. I was thinking of the case when the users know the pipe is PE80 but nothing more.

3nids:
thanks!

3nids:
Well, I think we should be coherent. If we have for others then we should have it for PE80 too.
Then, users are free to disable some of the materials (vl_active).
Do you pull a request for this?

tudorbarascu:
Yes.. right away. I will also change a data source in your qgs project as you have 1 or 2 tables still referring to qwat instead of qwat_od etc. (and I think it's not intentional)

PR #18, opened 2015-02-05
fix sql script as file was encoded as UTF-8 Unicode (with BOM) text

tudorbarascu:
sql was not working as  02_tr_valve_pipe.sql was encoded as "UTF-8 Unicode (with BOM) text".
When executing the create_schemas, '<feff>' was appearing out of nowhere in the code.

Script used: http://thegreyblog.blogspot.ro/2010/09/shell-script-to-find-and-remove-bom.html

Good morning and all the best.


3nids:
sorry for that, I edited the file in pgadmin3, that might be the reason!

PR #28, opened 2015-04-08
added valve handle geometry + view depicting valves with their handle as...

tudorbarascu:
Tried editing today with adding part; everything works as expected.
The only thing that's a little annoying is the usage of three layers (one for handle, one for valve and the line view..).

### DIFF
--- a/ordinary_data/08_valve/07_vw_valve_lines.sql
+++ b/ordinary_data/08_valve/07_vw_valve_lines.sql
@@ -1,0 +1,18 @@
+
+/*
+	qWat - QGIS Water Module
+	
+	SQL file :: valves_lines view
+*/
+
+
+DROP VIEW IF EXISTS qwat_od.vw_valve_lines;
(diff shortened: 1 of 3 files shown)


ordinary_data/08_valve/07_vw_valve_lines.sql
```diff
@@ -0,0 +1,18 @@
...
+	qWat - QGIS Water Module
+	
+	SQL file :: valves_lines view
+*/
+
+
+DROP VIEW IF EXISTS qwat_od.vw_valve_lines;
+CREATE OR REPLACE VIEW qwat_od.vw_valve_lines AS 
+ SELECT valve.id,
+    ST_MakeLine(valve.geometry_handle, valve.geometry) AS geometry
```
3nids: Is there no need to cast the geometry? 
i.e `::Geometry(LineString, 21781)`  ?

tudorbarascu: No need, but certainly better.
Thanks for observing.
Closing for repulling.

PR #40, opened 2015-06-09
added hydrant materials

tudorbarascu:
For increased consistency, Do you think we should move the hydrant_provider values to start with 71 let's say?


tudorbarascu:
ping  @3nids

3nids:
thanks for this
+1 for hydrant provider

Issue #32, opened 2015-06-03
Leaks improvement to designate repairing team


tudorbarascu:
Hi,

Regarding the leaks, we have a need to add the repairing team in the leaks data-model/ui. 
Are you interested in this feature so that I can make a pull request?
As always, any suggestions are welcomed.

All the best,
Tudor

3nids:
go for it.

I would suggest that we have a table of workers that might be available for specific works (leaks repairing for instance, and later valves checking, and so on)
They would have a date_entry and date_end.

So we don't have something specific for leaks.

thanks

haubourg:
Hi, @tudorbarascu  can we close that issue since it has been merged more than a year ago?

tudorbarascu:
Yes, it was left open as a reminder to discuss possible enhancements on maintenance, see https://github.com/qwat/qwat-data-model/pull/43

I had some ideas (lost my notes) at that time regarding maintenance. While waiting for the newcomers to join the discussion, somehow we passed on.

PR #43, opened 2015-06-15
added maintenance workers

tudorbarascu:
See: https://github.com/qwat/qwat-data-model/issues/32

I think we need to discuss it more as I think we can improve a lot.
Also, should we also make the maintenance more general in the data model?

Tudor


3nids:
indeed, it would require a bit of discussion.
I just would like to postpone this a bit if it's ok.
we'll have new comers in the group, and it would be nice to discuss this with them

tudorbarascu:
Good to have others joining in. We'll postpone it untill everyone is ready.

PR #46, opened 2015-06-18
Valve refactoring take 2

tudorbarascu:
Hello again,

I added the:
- handle_altitude_real
- handle_precision
- handle_precisionalti 
columns for the valve handle in order to avoid a possible problem where users could mistakenly think that the altitude_real, precision and precisionalti columns refer to the handle and not the actual valve-node.
After I get your ok on this will also make a pull for the valve ui. 
Also, I will need to create some check constraints so that in the case the handle_geometry is present --> non-null values in the handle precision columns.

Regarding valve types, I simply added the "gate valve" in the value list + filled english values.
However, I am proposing to also move the shut-off/arrêt and the "vent/jauge" valves from type to function. Thoughts on this?

It's also clear to me that the valves still need some refactoring and I'm waiting for the others qwat users to join in and make proposals.
I think that in the end we shall end up with a Qlist widget for valves_function as I believe there are some cases where we need to input multiple functions for a valve. I suppose we could even end up with a similar solution as for the pipe materials.

All the best!

### DIFF
--- a/ordinary_data/08_valve/01_od_valve.sql
+++ b/ordinary_data/08_valve/01_od_valve.sql
@@ -12,21 +12,24 @@ COMMENT ON TABLE qwat_od.valve IS 'Table for valve.';
 
 /* columns */
 ALTER TABLE qwat_od.valve ADD COLUMN identification    varchar(20) ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_distributor    integer not null      ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_type           integer not null ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_function       integer not null ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_status         integer not null ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_actuation      integer not null ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_precision      integer not null ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_precisionalti  integer not null ;
-ALTER TABLE qwat_od.valve ADD COLUMN fk_maintenance    integer[];
-ALTER TABLE qwat_od.valve ADD COLUMN fk_locationtype   integer[];
-ALTER TABLE qwat_od.valve ADD COLUMN diameter_nominal  varchar(10) ;
-ALTER TABLE qwat_od.valve ADD COLUMN year              smallint CHECK (year IS NULL OR year > 1800 AND year < 2100);
-ALTER TABLE qwat_od.valve ADD COLUMN closed            boolean       default false ;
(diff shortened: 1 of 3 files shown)


3nids:
Don't you consider we could have a single precision field for both the handle and the node?
I'm afraid it's getting too detailed.

What is a gate valve?

tudorbarascu:
It may look like it's crowding up but for networks that are not well determined (at least in Romania these problems exist) we can see the valve handle- head in the ground and constraint its position with accuracy. However, this doesn't mean the intersection between the valve and the pipe is right beneath, and we have many cases where the actual valve node is one meter on the left side or on the right side of the valve handle (if you're positioned parallel to the pipe). As we constrained the valve handle location, the valve-pipe node is in the buffer zone of 1 meter but as we don't know for sure.. it should have different precision.
If you think this situation is specific to Romania only I think we can keep it in the data model and add a special hidden tab in the UI that can be turned on/off with python. Or I can just keep these changes locally.

Thoughts?

For gate valve: https://en.wikipedia.org/wiki/Gate_valve
It's what you call, I believe,  "vanne à opercule".

tudorbarascu:
Hi Denis,
I am rebasing again to merge your commits, please add some comments to it so that I can adjust it also to your satisfaction.

3nids:
OK, I will merge the data model.
For the UI, yes please keep the changes locally until we have a good approach for taking vare of customization.

tudorbarascu:
Ok, how about my proposal of moving the shut-off/arrêt and the "vent/jauge" valves from type to function? Should I leave things like they are now?

3nids:
ok for moving them!

tudorbarascu:
It's good to merge now.
However, I noticed I didn't get the "Successuflly created all the schemas" echo at the end.
I checked it scarcely and it's on account of your changes.
Will look into it later on today if it's wanted or not.

All the best!

3nids:
thanks a lot.
I will complete by renaming fk_precision to fk_node_precision.
Also, I was thinking of keeping the 10'000 values for custom values.
I will modify the unknown / to be determined / other to 101,102,103.

I will test this afternoon the creation and will fix it.

thanks for your work!

tudorbarascu:
thanks also for your work,

I am in the process of adding the other/unknown etc. values to all the vl_* as we talked.
I will put them to 101, 102 and 103 then and leave the 10k values to custom ones then, as you suggested.

I will probably finish this weekend.

PS. please say if you want me to take care of what you said you'll do today (rename fk_precision to fk_node_precision and 101..103 for valves types and function) as I have some time now.

3nids:
it's ok, doing it right now!
I should commit a few things soon!

3nids:
after inspection, what I called "arrêt" where actually 1/4 turn valve.
so, in fact, this should stay in type.
update will follow soon ;)

PR #47, opened 2015-06-22
Installation buildings

tudorbarascu:
- added polygon_geometry for installation_valve chamber

How do you currently present the installation ground projection?
As I want to be able to describe correctly the setup, a polygon geometry would be needed.
The cover should also have the polygon geometry and should be in a parent-children relation with the corresponding installation. What do you think?

The following image depicts an usual valve chamber setup:

valves


3nids:
For this you're supposed to use qwat_od.installation_building. You have geometry_polygon in there. You can also set a type of installation (e.g. valve chamber). Then, you can also associate any specific installation (tank / pump / treatment / valve chamber etc.)

tudorbarascu:
It should be ok, I can't recall why I thought it wouldn't be enough. If I do, will let you know.

PR #50, opened 2015-06-29
Homogenize values

tudorbarascu:
Passed through all the tables.

I skipped the vl_visible table , if you think that one of the 101, 102 or 103 id's could be relevant there, please say so and I'll .. rebase accordingly.


3nids:
thanks!

PR #54, opened 2015-07-04
added no_valves & depth column for the installation chamber; renamed altitude_real to altitude, altitude_reference to object_reference

tudorbarascu:
Working on this, I noticed that for od_part, the altitude_real has a correspondance to the vl_altitude_reference through the fk_altitude_reference. 

I feel that altitude_real should always refer to the sea level reference. Have you thought otherwise?

If you think this is good, I will rename altitude_real to altitude for od_part at least.

I didn't had time to do the trigger for no_valves and the corresponding config table and as I will be away for few days I will do them as soon I get back.


3nids:
Indeed, it is with respect to sea level, altitude_reference determines where actually was measured the altitude ( apron / pipe / ground ).

The "_real" was meant by opposition to the altitude_dtm, maybe this is not really appropriate and should be renamed to altitude. Opinion?

tudorbarascu:
- both altitude or altitude_real are fine but I would keep altitude_real as the altitude referenced to the sea level
- as altitude_reference is not what its name suggests (to what reference is measured) but a reference to where the altitude was measured on the object I would rename it to something like altitude_measured_(at or on) or object_altitude_at or to what you suggest
- also, as we also want for some objects to know their depth we would need a new column. The column could have the name of altitude and have the corresponding  altitude_reference column as  to what reference is measured the altitude -certainly not the sea level as for that we would have the altitude_real. One caveat to this is that your users could mistake the new logic to the old logic as the altitude and altitude_reference would have a different meaning
Thanks,
Opinion?

tudorbarascu:
Good morning, 
I've rebased. Any opinions to my proposals?

3nids:
b) what about object_reference ? We could think it for other things than altitude only. That would make it better for c) which could simply name `depth`

a) therefore, I would vote for having simply `depth` and `altitude` (without _real)

tudorbarascu:
Sounds good to me. Will get onto it.
Further down the road I will probably need some suggestions for a corresponding "depth_reference" of the depth column. I will formulate better when I'm there.

tudorbarascu:
For the qwat_od.installation_pump we currently have altitude and altitude_real --> I will remove the altitude_real as it was probably a slipup.

3nids:
yep!

tudorbarascu:
It's good to go. It took less time than I thought. Will add the "depth" column to other objects later on when I'll pass over them. Thanks

3nids:
perfect, thanks

tudorbarascu:
Pleasure,
If you're not working on qwat.qgs and the UI's I can refit them to the data model and make a pull request.
I need your go here so that we don't overlap.

3nids:
ok, you can go for it!

PR #55, opened 2015-07-17
enhance materials table to be compatible with data driven props for diameter

tudorbarascu:


tudorbarascu:
I remembered why I needed so many cases for data driven properties.
The materials table had some rows with diameter completed for diameter only (text-string).
I filled the diameter_external for plastic materials and diameter_nominal for others.

For value 5652, PL 0.2, plastic, I put 20 mm but I think it may be something else.
As it's in plastic, it shouldn't be in inches, and as it's 0.2 it couldn't be 20cm--200mm, right?
Please advise so I can rebase or you can modify it later on.

Thanks

3nids:
I will look on Monday. 
Also,  the trigger for display name will need some attention to handle null.

tudorbarascu:
Morning. 
Will check out the trigger.
I didn't look at it as the values in display_name for diameter are looking as they should.

tudorbarascu:
Just checked and the trigger is handling NULL values in diameter since the https://github.com/qwat/qwat-data-model/commit/b8bc6af0d2257a6ab3e0c0701b794cc5f954fce4 commit

3nids:
perfect, thanks!

PR #56, opened 2015-07-19
temporary fix on fn_node_set_type so that pipes can be edited

tudorbarascu:
Although the model creation works, the fn_node_set_type didn't work and pipes couldn't be created anymore.
I noticed you had a TODO written and probably you have bigger fixed planned so I didn't try and do a more elegant fix.

Problems:
 PostGIS error while adding features: ERROR:  column reference "srid" is ambiguous 
    LINE 1: select srid           from geometry_columns where f_table_sc...

I changed the srid variable name to srid_var and changed the srid to 21781 in geometry::geometry(Point, srid).

PS. Before you changed the init scripts I was using this command to change the hardcoded values:
find . -type f -print0 | xargs -0 sed -i 's/21781/3844/g'
I didn't mind the hardcoded srid.

### DIFF
--- a/ordinary_data/nodes/fn_node_set_type.sql
+++ b/ordinary_data/nodes/fn_node_set_type.sql
@@ -28,7 +28,7 @@ $BODY$
 		stmt            text                     ;
 		keep_type       boolean          := false;
 		complement_col  varchar(50)      := ''   ;
-		srid            integer                  ;
+		srid_var        integer                  ;
 	BEGIN
 		/* determine if the node is under an object (hydrant, valve, etc.)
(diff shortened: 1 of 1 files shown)


3nids:
dd17bc6ba57e838ae0ea8cf59b647d1bfecea6bc should fix it.

tudorbarascu:
It's still not fixed:

Provider errors:
      PostGIS error while adding features: ERROR:  type modifiers must be simple constants or identifiers
    LINE 1: SELECT geometry::geometry(Point, qwat_sys.fn_setting_srid())...

3nids:
should be ok now!!!

PR #58, opened 2015-07-21
Steel materials

tudorbarascu:
Also, I think that the Steel/Acier pipe of 60 nominal diameter shouldn't exist in the pipe materials table as I couldn't find it in any standards. The equivalent that probably got that value written in the table is the Steel pipe of 50 nominal diameter that indeed has 60 as an approximate external diameter (so id 5258 should be replaced by 5257 for materials).
I think we should update the materials for pipe and delete it. Should I add this commit to this pull request?

All the best!


tudorbarascu:
ping @3nids

3nids:
I'm on it!

3nids:
remove stell 60 in a6fb959e6f6a9ab744e3092b3d60bace6b9b362b

3nids:
thanks for the input!!!

PR #61, opened 2015-09-25
remove non std. materials

tudorbarascu:
As sometimes people in production refer to the approximate
external diameter (for steep pipes) instead of the DN,
this commit ensures the correct materials can be corelated
with their DN.

Also removed pipe material with steel DN75, DN70, DN120, DN 180

I couldn't find them in any standard

For migration matching in case anyone has saved pipes with the "wrong" materials:

DN120     5103 should go into 5104      (DN125)
DN180     5107 should go into 5106      (DN175)
DN70       5117 should go into 5116      (DN65, DE 73)
DN75       5118 should go into 5116      (DN65, DE 73)

Something like this should work:
`update qwat_od.pipe set fk_material = 5116 where fk_material in (5117, 5118);`


tudorbarascu:
You can look at http://www.wosupply.com/pdf/1_General_Products_82511_113.pdf or any other resource with steel dimensions standards, and the DN standards are:
- 20
- 25
- 32
- 40
- 50
- 65
- 80
- 90
- 100
- 125
- 150
- 175
- 200
- 250
- 275
- 300
- 350
- 400
- 450
- 500
- 550
- 600

tudorbarascu:
I also added a commit for french translation of cubic stones,  `pierres cubiques`, please say if it's a correct one.

3nids:
For cubic stones, it's just "pavés".

I will check for the DN with my colleagues (in a few hours) and merge this.

tudorbarascu:
I've rebased it with the 'pavés' value

3nids:
I just need a bit of time before merging this one to clean my data.
Hope to do it today.

tudorbarascu:
do you think I should split this pull between the material removals and the other commits so that they won't wait in queue?

3nids:
sorry I couldn't sort this out today, I will take care of this when I come back (2 weeks).

tudorbarascu:
No worries :). Everything is as it should be as the other commits got in through the other pull request.

3nids:
thanks!

PR #64, opened 2015-10-04
fix for new inserts for 3D geometry

tudorbarascu:
The triggers that make the 2d to 3d translation don't get to run
due to typemod constraints that are run before the geometry is fixed.
This enables the old behaviour of storing geometry with check constraints.

Followup https://github.com/qwat/qwat-data-model/commit/78e9dd183fae78ec39dbf86a404aad5d727a5e66

### DIFF
--- a/ordinary_data/nodes/od_node.sql
+++ b/ordinary_data/nodes/od_node.sql
@@ -40,9 +40,9 @@ ALTER TABLE qwat_od.node ADD COLUMN _pipe_schema_visible boolean default false;
 ALTER TABLE qwat_od.node ADD CONSTRAINT chk_node_altitude_obj_ref CHECK (fk_object_reference IS NOT NULL OR altitude IS NULL );
 
 /* GEOMETRY */
-ALTER TABLE qwat_od.node ADD COLUMN geometry geometry('POINTZ',:SRID);
-ALTER TABLE qwat_od.node ADD COLUMN geometry_alt1 geometry('POINTZ',:SRID);
-ALTER TABLE qwat_od.node ADD COLUMN geometry_alt2 geometry('POINTZ',:SRID);
+SELECT AddGeometryColumn ('qwat_od','node','geometry',:SRID,'POINT',3, false);
+SELECT AddGeometryColumn ('qwat_od','node','geometry_alt1',:SRID,'POINT',3, false);
+SELECT AddGeometryColumn ('qwat_od','node','geometry_alt2',:SRID,'POINT',3, false);
 CREATE INDEX node_geoidx ON qwat_od.node USING GIST ( geometry );
 CREATE INDEX node_geoidx_alt1 ON qwat_od.node USING GIST ( geometry_alt1 );
 CREATE INDEX node_geoidx_alt2 ON qwat_od.node USING GIST ( geometry_alt2 );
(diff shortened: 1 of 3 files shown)


3nids:
Hi Tudor 
Thanks a lot for digging in and finding the issue. 
For a first patch, I would prefer to modify the insert and update triggers on qwat_od.vw_node_element,  I think it will solve the problem. I will test tomorrow. 

Then there will be a problem for the pipes, a new view might be required.

tudorbarascu:
Hi Denis,
Nice to be able to help.
I've rebased so that only the geometry column creation is touched.

3nids:
this is not working as expected, I am working on a solution

tudorbarascu:
This should just enables the triggers to run before the checks.
Were you expecting something else?

3nids:
Have you tested inserting features ?
I still got the error.

I am almost done with the new solution, I am at a meeting now, but will finish this afternoon. I just use a force2d in the view for now and force back to 3d when inserting.

tudorbarascu:
Yes, I tested for nodes and for pipes and it worked!
For pipes I had to insert the force3d line in the trigger in order to work but basically, it worked from QGIS and from a manual insert using 2d points.

tudorbarascu:
Just recreated the database model using master and I can add nodes just fine using QGIS.
Although I have filled the altitude column with 300, the geometry z gets saved with 0 currently, but I think this falls on the triggers.
Moreover, by reworking the `qwat_od.ft_node_geom()` so that it has `NEW.geometry := ST_SetSRID( ST_MakePoint( 400000, 400000,300), 21781);` the point gets saved as dictated by the trigger.

So, It's working as expected. Were you expecting something else?

3nids:
When I tried I couldn't save a point and got the same postgis error. I did probably something wrong. 
I have almost finished the views with force 2d/3d.
I'll let you know.

Best wishes 
Denis

3nids:
Here is the error I get when trying to add an hydrant
```PostGIS error while adding features: ERROR:  new row for relation "node" violates check constraint "enforce_dims_geometry"
    DETAIL:  Failing row contains (30477, null, null, null, null, null, null, null, null, null, null, null, null, null, 0101000020155500007517469B62262141ACBF64975CCA0141, null, null)```

3nids:
I think the problem here is that there is no geometry_alt1/2 and there is a problem with null geometries.

tudorbarascu:
Good morning,
Is this on the latest model?
If not, at what commit, as I thought you reverted the check constraints mode.

3nids:
I checked out 0cb54e477acfc87cafa04ead8c732af98fb962cd, so just after your PR

tudorbarascu:
I'm going to look at it now.
I tested only adding nodes and pipes on the latest qwat.qgs project.

3nids:
yeah, I suppose it works since there is no NULL geometry occurences in this procedure.

tudorbarascu:
I see what you mean.
`
insert into qwat_od.distributor (name) select 'Test Distributor';
insert into qwat_od.vw_element_hydrant
(fk_distributor, fk_status, fk_material, fk_precision, fk_precisionalti, geometry)
SELECT 1, 101,101, 101, 101, ST_Force3D(ST_SetSRID( ST_MakePoint( 400000, 400000), 21781));
`
This works as I used force3d, without it, I get the error you mentioned.
I am busy right now but I will investigate this afternoon/evening.

I believe though it is on account of the views/triggers.
All the best Denis.
Thanks for your work.

3nids:
thanks for digging in.
I will fix that!!!!

3nids:
done for the nodes 47ab0b424ca0adc02c328fbea8e94f6f5fef34aa

tudorbarascu:
Good morning,
Do you still need me to investigate saving the hydrants?

PR #66, opened 2015-10-09
add 'on top' of object altitude reference value

tudorbarascu:


3nids:
on top of what? ;) isn't it the ground?

tudorbarascu:
It's for cases where objects are not fully underground.
A tank may not be completely in the ground, having it's base buried and the top above the ground.
In this case, easiest way of doing altitude measuring is on top of it as the ground level doesn't have relevance.

3nids:
Well in case of reservoir, the altitue is usually the apron.
I just wonder if it makes sense to add these distinctions.
Shouldn't we use surveypoints for this?
What do you think?

tudorbarascu:
Yes, but you're thinking about the good cases where data is good and all the records are kept.
We have some old reservoirs where we don't have complete records (no apron) but we'd still like to present in the form whatever data we have.
I can keep this specific to our case but as we can query the values of the layer in the QGIS project, I thought it would be good to have them in the model for flexibility.
Your choice :).

Not related, do you still have work to do on the model and project as I would like to test/migrate to it?
All the best.

3nids:
ok, good for me.

yes, the work on the model should be complete. I did quick tests and it was ok.
there are still some work to do for the surveypoints, but no big deal.

PR #70, opened 2015-10-16
Fixed valve with handle view/demo distributor/removed node_altitude

tudorbarascu:
node_altitude values should go into altitude when migrating

### DIFF
--- a/ordinary_data/demo.sql
+++ b/ordinary_data/demo.sql
@@ -1,0 +1,5 @@
+/* qWat - Demo data */
+
+INSERT INTO qwat_od.distributor (name) VALUES ('Demo Distributor');
+
+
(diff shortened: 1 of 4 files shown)


3nids:
Thanks a lot. Please put the demo data in a distinct file and it'll be perfect !

tudorbarascu:
Thanks also! I put the demo.sql in the ordinary_data dir. Hope it's ok.

3nids:
Perfect, you might want to add it to the init file too, with maybe an option similarly to the create roles option.
Otherwise I'll do it when I come back.

tudorbarascu:
I'm busy now with migrating/testing everything, will do it afterwards ;).

PR #83, opened 2015-12-15
fix altitude trigger

tudorbarascu:
This is to fix the insert case with 2d geometry and null altitude when:
``insert into qwat_od.cover
(fk_distributor, fk_installation, fk_status, geometry)
SELECT 1, 1, 101, ST_SetSRID( ST_MakePoint( 400001, 400001), 21781);``

outputs:

``ERROR:  new row for relation "cover" violates check constraint "enforce_dims_geometry"
DETAIL:  Failing row contains (2, null, 1, 101, null, 1, null, null, t, null, null, 01010000201555000000000000046A184100000000046A1841, null, 1, null, null, null, null, 1, null, null, null, null)``


tudorbarascu:
@3nids Any thoughts on this? :)

3nids:
yes, looking today

PR #85, opened 2015-12-17
Moving year_end before year_rehabilitation

schoeller:
During installation the execution throws the error:

<code>psql:ordinary_data/pipe/od_pipe.sql:26: FEHLER:  Spalte "year_end" existiert nicht</code>

This seems due to that <code>year_end</code> should be called first.


3nids:
I am not getting such error, and I don't see how this could be solved by this.

schoeller:
My bad. Sorry

PR #90, opened 2016-01-26
Add a pipe vertex when a node intersects a pipe

### LINKED ISSUE #88: Automatically add a vertex when adding an object on pipe
When adding a new object on a pipe (valve, other punctual element), a new node should be automatically created using a trigger.

mhugo:
Should fix #88 
The unit tests have also been changed a bit


3nids:
Good job for the tests!
I would be in favor of going with a tolerance, can you update this?

Who is in charge of merging? Am I?

mhugo:
Ok for the tolerance, it would require to also update the node creation function. I'll have a look.
I will merge (this was just to have your feedback) and continue to work on it

mhugo:
Tolerance added in 586894f0

PR #99, opened 2016-02-12
fix altitude trigger

tudorbarascu:
This is to fix the insert case with 2d geometry and null altitude when:
``INSERT INTO qwat_od.cover
(fk_distributor, fk_installation, fk_status, geometry)
SELECT 1, 1, 101, ST_SetSRID( ST_MakePoint( 400001, 400001), 21781);``

outputs:

``ERROR:  new row for relation "cover" violates check constraint "enforce_dims_geometry"
DETAIL:  Failing row contains (2, null, 1, 101, null, 1, null, null, t, null, null, 01010000201555000000000000046A184100000000046A1841, null, 1, null, null, null, null, 1, null, null, null, null)``

### DIFF
--- a/ordinary_data/generic_functions/ft_geom3d_altitude.sql
+++ b/ordinary_data/generic_functions/ft_geom3d_altitude.sql
@@ -5,13 +5,20 @@
 CREATE OR REPLACE FUNCTION qwat_od.ft_geom3d_altitude() RETURNS TRIGGER AS
 $BODY$
 	BEGIN
-		-- altitude is prioritary on Z value of the geometry (if both changed, only altitude is taken into account)
-		IF NEW.altitude IS NULL THEN
-			NEW.altitude := NULLIF( ST_Z(NEW.geometry), 0.0); -- 0 is the NULL value
-		END IF;
-		IF 	NEW.altitude IS NULL     AND ST_Z(NEW.geometry) <> 0.0 OR
-			NEW.altitude IS NOT NULL AND ( ST_Z(NEW.geometry) IS NULL OR ST_Z(NEW.geometry) <> NEW.altitude ) THEN
-				NEW.geometry := ST_SetSRID( ST_MakePoint( ST_X(NEW.geometry), ST_Y(NEW.geometry), COALESCE(NEW.altitude,0) ), ST_SRID(NEW.geometry) );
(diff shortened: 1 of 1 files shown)


3nids:
@tudorbarascu this should be fixed now, correct?

tudorbarascu:
I'll retest and close it if that's the case. As it's been some time and no feed I was lazy enough and did not retest each time a new commit arrived. Thanks for the heads up!

3nids:
yes, I understand :)

tudorbarascu:
@3nids I just tested and it's exactly the same error.
The only thing that can be improved is to set the z to `-9999` or `-99999` but as we haven't established a consensus on that..

3nids:
I'm not a huge fan of making the altitude trigger more complex to fix a 2d/3d problem.
Now that QGIS works fine with 3d, is this really a matter?
If you go via SQL, then it's not big deal to add a ST_Force3d.
What do you think?

tudorbarascu:
Indeed, it wouldn't matter if it would work.
For me it doesn't work in a simple case, so it's a show stopper.

With latest Qgis Master, latest qwat model latest qwat.qgs project:
- add an installation (ouvrage)
- adding a cover on that installation throws:

``Could not commit changes to layer couvercles
 Errors: ERROR: 1 feature(s) not added.
  Provider errors:
      PostGIS error while adding features: ERROR:  new row for relation "cover" violates check constraint "enforce_dims_geometry"
    DETAIL:  Failing row contains (2, null, 1, 1301, 9801, 1, 2013, null, f, null, null, 01010000201555000061C6A8AFBFF020417545E462B7BD0041, null, 1, null, null, null, null, 1, null, null, null, null).``

I'm very curious if this works for you and if so, what are you doing differently.
I'm doing everything from QGIS, so.. no direct SQL.

Thanks for your time.

3nids:
should be fixed in last qwat project with QGIS master https://github.com/qwat/QWAT/commit/28c04aced4875979bc0f7dcae88ab5fe83efe5ff

3nids:
I also switched back column creation to Postgis 2+ style 980ed2f910343c2bd45fdb3aa719b4dcea873dba

tudorbarascu:
It's solved now! :+1:  Thanks and all the best

PR #101, opened 2016-02-18
Disable the creation of pipe vertex when doing a pipe creation

mhugo:
@3nids is it this kind of change you had in mind ?


3nids:
well yes. I was hoping for something cleaner than deactivating/enabling triggers, but I have nothing better to propose.

I would just rephrase the argument of the function to `deactivate_node_add_pipe_vertex boolean = False`

ok to merge for me

mhugo:
Idem, nothing better than that for now ...

PR #105, opened 2016-02-23
2d_length removal

tudorbarascu:


ordinary_data/views/schematic/vw_pipe_schema.sql
```diff
@@ -17,7 +17,6 @@ CREATE OR REPLACE VIEW qwat_od.vw_pipe_schema_visibleitems AS
 		pipe.id,
 		pipe.fk_parent,
 		pipe.fk_material,
-		pipe._length2d,
```
3nids: you need to recompute the length here with `ST_Length(geometry) AS _length2d`

otherwise we won't have such info since we only keep geometry_alt2

ordinary_data/views/schematic/vw_pipe_schema.sql
```diff
@@ -46,14 +45,13 @@ CREATE OR REPLACE RULE rl_pipe_delete_alternative AS
 		
 
 CREATE OR REPLACE VIEW qwat_od.vw_pipe_schema_merged AS	
-WITH RECURSIVE pipe_find_parent(/*path,*/ depth_level, id, groupid, geometry, _length2d, _length3d, tunnel_or_bridge, _valve_count, _valve_closed) AS (
```
3nids: hence, here and after we can keep `_length2d`

3nids:
perfect, thanks

PR #115, opened 2016-03-25
add unique constraint on pipe geometry

tudorbarascu:
As I just found some duplicate geometries in the pipe data that I work with I'm enforcing a unique constraint on the geometry column.
I think this should also be here (upstream for me).
@3nids & others: How does this sound to you?


tudorbarascu:
It seems this also touches some pipes that start from A to B but have different geometries.
Closing until I'll find a better solution.

Issue #108, opened 2016-03-10
moving node (installation, valve, etc) away from a pipe end will fail


3nids:
If you move a node from an end of pipe (a valve or an insallation for instance), this will cause an error.
This is another side effect of the automation of vertex addition to pipe.

I created a test to reproduce the error
https://github.com/qwat/qwat-data-model/blob/master/tests/test_move_node_end_pipe.sql

@mhugo can you or someone else from Oslandia have a look to this?

3nids:
ERROR:  tuple to be updated was already modified by an operation triggered by the current command
    HINT:  Consider using an AFTER trigger instead of a BEFORE trigger to propagate changes to other rows.
    CONTEXT:  SQL statement "UPDATE qwat_od.pipe SET	fk_node_a = qwat_od.fn_node_create(ST_StartPoint(geometry)) WHERE fk_node_a = OLD.id"
    PL/pgSQL function qwat_od.ft_pipe_node_moved() line 5 at SQL statement
    SQL statement "UPDATE qwat_od.node SET ......

3nids:
I believe the problem is that 2 triggers are run which will update the same pipe in this particular case: `qwat_od.ft_node_add_pipe_vertex` and `qwat_od.ft_pipe_node_moved` will be triggered on geometry update on `qwat_od.node`, although both triggers are `AFTER`

ponceta:
Seems that Sébastien found it yesterday too. :

![image](https://cloud.githubusercontent.com/assets/8614223/13675637/de354352-e6e2-11e5-983b-36e065bf9ecd.png)

vpicavet:
Do you have a SQL script that triggers the error ?

3nids:
Link in the description ;)

vpicavet:
indeed ;-)
Will add to tasks list

sylvainbeo:
The SQL test script is working here. No errors. I tried it a lot. Still works.
I'm going to have a look at the triggers anyway.

3nids:
I just tried again on a fresh install, just run this test and got the same error....
```
ERROR:  tuple to be updated was already modified by an operation triggered by the current command
HINT:  Consider using an AFTER trigger instead of a BEFORE trigger to propagate changes to other rows.
CONTEXT:  SQL statement "UPDATE qwat_od.pipe SET	fk_node_a = qwat_od.fn_node_create(ST_StartPoint(geometry)) WHERE fk_node_a = OLD.id"
[...]

PR #118, opened 2016-04-21
moving node (installation, valve, etc) away from a pipe end will fail #108

sylvainbeo:
I think i solved the problem.
No an easy one.

Denis, please have a look, test, and give me feedback, so we can validate or work again on this one.


ordinary_data/nodes/od_node.sql
```diff
@@ -119,16 +119,28 @@ COMMENT ON TRIGGER tr_node_add_pipe_vertex_update ON qwat_od.node IS 'Trigger: u
...
+        IF start_geom IS NOT NULL THEN
+            new_node_a := qwat_od.fn_node_create(start_geom);
+        END IF;
+        IF end_geom IS NOT NULL THEN
+            new_node_b := qwat_od.fn_node_create(end_geom);
+        END IF;
+        UPDATE qwat_od.pipe SET fk_node_a = new_node_a WHERE fk_node_a = OLD.id;
+        UPDATE qwat_od.pipe SET fk_node_b = new_node_b WHERE fk_node_b = OLD.id;
+        RETURN NEW;
+    END;
```
3nids: what happens if start_geom is NULL? then the pipe gets a NULL node?

It seems to work (the test succeed), but I have trouble at understanding the idea (too close to the weekend ;) )...
can you give me a short explanation?
thanks!

sylvainbeo: If start_geom is NULL then the following UPDATE will fail (because there's no fk_node_a = OLD.id). In fact we could add on more test to avoid the UPDATE in that case.

3nids: yes, that would be perfect.
To merge, can you write a few line of comments and enable corresponding test?

3nids:
ping @sylvainbeo

3nids:
@sylvainbeo @vpicavet any chance to get this through?
this issue is 2 months old...

sylvainbeo:
Sorry for the delay. I am very busy.  it Will be done tonight or tomorrow

sylvainbeo:
Done.

3nids:
thanks, but the test is not yet enabled. it should be added in tests.sh

sylvainbeo:
Oh yes, sorry. Corrected. I reorganized the test list in the bash script for more clarity.

3nids:
thanks a lot!

PR #121, opened 2016-06-06
add functionnal to vl status (meaning the object is ready to be used).

3nids:
this is useful to allow finer printing (print inactive but functional objects)

testing update process at the same time


3nids:
@sylvainbeo it seems the migration test only apply a single delta (1.0.0) and does not play the included delta here (1.0.1). Any idea?

sylvainbeo:
The travis feedback show us that it apply both : 

Processing delta_1.0.0_12092016.sql, num version = 1
Processing delta_1.0.1_06062016.sql, num version = 1

sylvainbeo:
By the way, the diff between the 2 models seem to be identical :

<     status.functional AS status_functional,
4039a4039
>     status.functional AS status_functional,

4295d4294
<     status.functional AS status_functional,
4296a4296
>     status.functional AS status_functional,
4585d4584

I don't see any differences...unless spaces or tabs  ?

3nids:
I was looking at the first entry for Travis
https://travis-ci.org/qwat/qwat-data-model/builds/179428840#L2267

What are actually the 2 lines (pr vs push)?

I don't understand what's going on here, some support would be appreciated :)

sylvainbeo:
There is a builds for the PR and for the branch push. travis works that way.
So, it may have differences between the 2 since the PR may have been done at a time, and the lastest push at anothe time.
So we have to focus on the push one. The PR will pass then.

3nids:
The problem is in the ordering of the columns. Alter table adds the column at the end, hence the export views created automatically (joining all fields of joined tables).

There is [no proper workaround](https://wiki.postgresql.org/wiki/Alter_column_position) for this.

What do you think about using `view_column_usage` and `view_table_usage`, since we could order the columns alphabetically? (I am not sure this would work as expected, I need further reading in that case).

sylvainbeo:
That's a big problem indeed. 
Or maybe, the integration process sould not verify the automatic generated views ? In fact, they are created with a script, hence if we check the tables, and the tables are identical, then the views will be identical.

3nids:
good point, will try this

3nids:
thanks a lot @sylvainbeo

PR #128, opened 2016-10-05
pipe year columns <= current year + column comments

tudorbarascu:
Reworked constraints on year, year_rehabilitation and year_end columns
constraints so that the values are smaller that the current year.
I also added comments on the constraints and on the columns.

Is there any reason that the year would be bigger than the current_year and smaller than 2100.

I am thinking about having a convention so that pipes that are in a project (with `fk_status` 1306 - which means they are in a project) could have a `year` value that is bigger than the current year and that it should represent the year when the pipe should be installed. I don't know yet if this would serve anyone, but I just thought of bringing it out as a possibility.

I am unsure of my English when describing the year_end column **.. year when the pipe was shut down**. Maybe someone can suggest a better expression than **shut down**.

Thoughts?


haubourg:
@3nids Hi Denis, can we merge that? any objection ?
Régis

3nids:
> I am thinking about having a convention so that pipes that are in a project (with fk_status 1306 - which means they are in a project) could have a year value that is bigger than the current year and that it should represent the year when the pipe should be installed. I don't know yet if this would serve anyone, but I just thought of bringing it out as a possibility.

indeed we should keep such flexibility. We are supposed to have such things in place.

tudorbarascu:
made the check constraint to consider the fk_status = 1306 case.

Do you think that we would be better with a trigger so that we can benefit from a better message on why the check failed?

3nids:
Well, I wouldn't go that far...

For project, I would set up the range to be now()-5 years up to now() + 50 years.

You also need to provide a delta file for any new change in the model. See https://github.com/qwat/qwat-data-model/pull/121/files for instance.

tudorbarascu:
@3nids @haubourg  I tried solving this using check constraints but I am unable to cover all the cases with the built in functions. If anyone wants to step up.. :+1: ?
I can do it of course with a function/trigger.

lbartoletti:
@tudorbarascu @haubourg still a WIP? Can be checked and merged for next release?

tudorbarascu:
@lbartoletti working on it!

tudorbarascu:
@lbartoletti Done!

PR #140, opened 2016-12-13
Building trigger in one shot (no more hierarchy)

sylvainbeo:

### DIFF
--- a/ordinary_data/views/inheritance/od_all_inheritance.py
+++ b/ordinary_data/views/inheritance/od_all_inheritance.py
@@ -1,0 +1,196 @@
+#!/usr/bin/env python
+
+import imp
+import os
+import sys
+
+pgiv = imp.load_source('PGInheritanceViewRecursive', os.path.join(os.path.dirname(__file__), '../../../metaproject/postgresql/pg_inheritance_view/pg_inheritance_view_recursive.py'))
+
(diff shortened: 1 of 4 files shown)


3nids:
views are not renamed (i.e. no need for new version?)

sylvainbeo:
Sorry did not see your message. Which views  ?

3nids:
I mean shall we see this as a model update and provide a dummy delta (with just the model increment) ? How do you see it?

sylvainbeo:
Ah. That's a good question.
Maybe it's best to add indeed an empty delta. At least to indicate it's a new version of the model. I do it.

3nids:
remove the local script?

sylvainbeo:
Ok travis is fine with the metaproject (master).

sylvainbeo:
The qwat-data-model needs to be updated to have the metaproject Master in sub repository.

3nids:
what do you mean? if you checkout to the proper commit in this branch (lvl_triggers_recursive), then master will be automatically updated when this PR is merged.
Am I missing something?

sylvainbeo:
Ok. I'm not very familiar with sub-git dependencies vs PR. Anyway, if that's not the case, it wil be easy to do it afterwards.

PR #156, opened 2017-03-21
printmap add scale field

3nids:
release as 1.2.2


haubourg:
Hi @3nids , there is a conflict in 1.2.2 version number with my pending request (still with failing tests though, @sylvainbeo is trying to check what's failing again)
Could you use version 1.2.3 ?

haubourg:
mmm I see you have the same emty diff message I have. @sylvainbeo , can you throw an eye?

3nids:
no big deal for the version number, but they should be merged in the proper order.
this one is quite straightforward...

is there any update for the db conformity checker?

sylvainbeo:
I'm currently checking this issue.

sylvainbeo:
@3nids The problem here is that "printmap.print_scale" is used in a view. And as we know, order in view is not very accurate. I'm checking.

sylvainbeo:
@3nids You don't add the print_scale column in the same order  than you do in the normal script.

sylvainbeo:
@3nids  You should add the "print_scale" column add the end. The fn_label.sql script add automatically some columns to the printmap table...this the problem.

3nids:
@sylvainbeo I thought you took care of discarding field order conformity. Didn't you?

sylvainbeo:
@3nids Not in that way. We did it with the auto generated views. But this is a different case. 

Postgres is recording the view with fields in order of tables. Impossible to check the conformity of the view if the fields are in different order.

haubourg:
That sounds reasonable, comparing DDL structure with different orders can be tricky. Having different QWAT installs with the same model version but different column order is also not quite right.

3nids:
thanks for the feedback, will try on Monday by dynamically rewriting the view 

@sylvainbeo although it is weird: I did not change any view, so there shouldn't be any change. And if it's an auto generated view, it should be handled as you said.

@haubourg we have no choice, you cannot change column order and it is not right to add them at the end.

sylvainbeo:
@3nids "I did not change any view, so there shouldn't be any change": no, but you changed the table structure. Thus the view is affected. Did i misunderstand your question ?

One other way to solve that kind of problem is to be more specific in the view declaration.

ordinary_data/views/export/vw_printmap.sql

Instead of 
CREATE VIEW qwat_od.vw_printmap AS
        SELECT
                printmap.*,
do

CREATE VIEW qwat_od.vw_printmap AS
        SELECT
                printmap.field1,
                printmap.field2,
               ...

But we have to think of that each time we add a new field...not very efficient.

PR #157, opened 2017-03-21
Better doc

haubourg:
Adds a better web based documentation for the data model. Generated using schemaSpy.
It is also browsable directly using rawgit so it would be nice to update it when data model changes.


haubourg:
@sylvainbeo @3nids This PR adds 48 Mo of html and png files. Is that OK with you pushing that on github ? 
On one hand, this is nice to be able to keep a versioned data model, on the other hand, this is heavy for git and we could just host that on the website somewhere. 
Opinions?

sylvainbeo:
Well, that's very nice. Thank you for that. Indeed it's very good to navigate thru the model, even if i'm not sure it will be clearer (see the relationship tab).
Anyway, i'm ok for merging.

3nids:
fine with me too!

haubourg:
@sylvainbeo agreed, schemaSpy is better than autodoc but not yet as good as enterprise Architect yet :)
The good open source tool doesn't exist yet, still searching the holy grrrail!

PR #158, opened 2017-03-22
[fix qwat/QWAT/#174] on delete cascade for subscriber and meter refer…

3nids:
…ences, on delete set null for pipe.fk_parent

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 
 # run on Travis container-based infrastructure
+dist: trusty
 sudo: false
 
 git:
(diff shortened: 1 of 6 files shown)


3nids:
see qwat/QWAT/#174

3nids:
@sylvainbeo would need a bit of help here too.

I am trying to change constraint actions. 

image

Any idea?

sylvainbeo:
Again, sounds like a bad ORDER in the check SQL.

Please edit you test_migration.sql,

Line 104 is : ORDER BY p.relname, t.tgname
Replace with : ORDER BY p.relname, t.tgname, pp.prosrc

Commit and push.

3nids:
will try on monday thanks for the feedback.

3nids:
@sylvainbeo apparently it doesn't solve the issue

sylvainbeo:
@3nids Seems fixed now.

3nids:
@sylvainbeo perfect, thanks a lot!

PR #179, opened 2017-07-03
extend ro values for hydrant_output and models + comment the hydrant_models tables

tudorbarascu:

PR #180, opened 2017-07-04
remove FALSE default for hydrant underground and marked

tudorbarascu:
The hydrant ``underground`` and ``marked`` columns are currently ``NOT NULL`` and with False as default.
Now that I look at them it doesn't feel right to have them default to anything as we should leave this to the user/editors to fill the possibilities.
I prefer NULL values instead of automatically filled ones that come with the uncertainty whether the editor actually chose the value or not.

Thoughts?


3nids:
Well, in practice about 99.50% of them are not underground, so I feel it's more practical.
Although this could be done on the client side, as it would let the administrator decide the default value.

So +0 on my side.

tudorbarascu:
@3nids Thanks for your thoughts.
Indeed, I was thinking of leaving this in the hands of the administrator, client side.
If we're going to use a default on the database side I guess we should document this so that normal users aren't baffled about the automatic filling database side.

Any opinions about the ``marked`` part?

3nids:
same, ok for me to merge.
@haubourg any objection to merge?

haubourg:
From a user point of view, I have no strong opinion on that.
From release management, we need to build a SQL patch in update with each commit, if we don't do so, it's going to be hard to reconstruct one later, and Travis tests won't work (when fixed).

tudorbarascu:
@haubourg  I was waiting for #179 to know the model number in the delta.

lbartoletti:
@tudorbarascu can you update your PR for next release, please?

tudorbarascu:
@lbartoletti This can be merged

lbartoletti:
Thanks @tudorbarascu !

lbartoletti:
target for release 1.3.3 #250

PR #181, opened 2017-07-04
[travis] Update to postgres 9.6

m-kuhn:


haubourg:
Hi, 
it seems we are not alone to face issues since 12 days:
https://github.com/travis-ci/travis-ci/issues/6972

adding this to config seems to solve the issue but costs 22 seconds of build time:
````
  apt:
    packages:
    - postgresql-9.6-postgis-2.3

````

**Comment by m-kuhn** [comment]
Thanks @haubourg , trying this now

**Comment by haubourg** [comment]
Thanks @m-kuhn 
Merged, we now are back to the authentication issue @sylvainbeo

**Comment by m-kuhn** [comment]
In case you have special characters ("special" in bash context) in the password, have a look at this instruction:

https://github.com/opengisch/qfieldsync/blob/master/.travis.yml#L34

**Comment by haubourg** [comment]
You mean adding double quotes around secure encrypted key would do?

**Comment by m-kuhn** [comment]
escaped single quotes + single quotes is what I needed to do.

But without having your password at hand it's very hard to give any guarantees :stuck_out_tongue_winking_eye:

**Comment by haubourg** [comment]
Oh, before encrypting password then.. I also saw double quote around the key in your travis example. 
@sylvainbeo we need you then !

**Comment by sylvainbeo** [comment]
I'm trying that.

**Comment by sylvainbeo** [comment]
Mmm, in our case we don't encrypt a password but a token, with no special char...

**Comment by sylvainbeo** [comment]
travis-encrypt -r qwat/qwat-data-sample GH_TOKEN=[xxxxxxxxxxx]

**Comment by m-kuhn** [comment]
then it shouldn't be an issue...

the same commands executed locally succeed?

**Comment by m-kuhn** [comment]
```
export GH_TOKEN=[xxxxxx]
./scripts/deploy.sh # or whatever it is for you
```

sylvainbeo:
We don't need to do the export. COnfig is directly in .travis.yml

sylvainbeo:
@m-kuhn Do you well understand the travis-encrypt command ? I mean, each time we execute it, it override the last generated public token ?

m-kuhn:
The "clear-text" token was generated once.

Everytime you encrypt it, it generates an encrypted version for this token. I think these encrypted versions can be used interchangeably.

sylvainbeo:
Ok, that's what i thought

Issue #80, opened 2015-11-30
cover enhancements/proposals


tudorbarascu:
Hi,
I was thinking of:
- adding a `fk_installation` cover so that we can attach covers to installations.
- for rectangular covers, it would be interesting of also having width, length, and rotation in order to have a rectangle symbol matching the actual cover projection
- I am inclined a little to ditch the node from covers. Although it isn't a real bother, it also doesn't feel right to have it as node as we have nodes for installations and the cover should be an installation child. In QGEP as cover is somewhat the main object, it makes sense.

Thoughts?

tudorbarascu:
ping @3nids

3nids:
covers do not inherit nodes anymore, it makes sense.
I also added fk_installation

For the rest, wouldn't it be easier to have a second geometry polygon ?

tudorbarascu:
Thanks.

A second polygon as geometry would also suffice.
I made the above proposal as I am a little uneasy with the current work-flow of adding a secondary geometry and I wanted to grasp your opinion on it.

My reasoning is that working with circle - diameter and rectangular (length, width + rotation) would actually save time when adding new covers with area presenting + decreasing storage needs.

I was also wondering about the existence of covers with shape different than that of a circle and rectangle. If so, we should drop the circular (boolean) and add a fk_form column for future proofing.
We of course have covers for hydrants that are oval shaped but I don't know if there will ever be a need to describe that.

tudorbarascu:
Actually, to my knowledge, we only have square/circle shaped covers.
So, circular (boolean) with dimension and rotation would also be an option.

tudorbarascu:
Regarding the ``circular`` column (with default ``true`` db side), do you think that we should change its type from boolean to integer (fk_form)?
This way we would have the unknown, to be determined .. values and we would also cover the rectangular form.
Indeed, if it's not circular, it should be rectangular.. so boolean seem to work.
However, the current process of adding a rectangular cover is too ..complex in that we add a cover, and as the default is true, we need to save it in the database and then uncheck the checkbox...

I will do a pull request if you concur.
Thanks

tudorbarascu:
ping @3nids

tudorbarascu:
Hello, while migrating I stumbled again on covers and the data model currently fits the bill but I think we can extend the covers to also have fk_precision, fk_precision_alti and  fk_object_reference (like before).
We could drop the NOT NULL on those if anybody is bothered by it.

Thoughts?

[...]

PR #188, opened 2017-07-14
add fk_precision, fk_precisionalti and fk_object_reference to cover

tudorbarascu:
See #80 

I'll make the delta file after the CI is working again.

### DIFF
--- a/ordinary_data/installation/od_cover.sql
+++ b/ordinary_data/installation/od_cover.sql
@@ -15,6 +15,9 @@ ALTER TABLE qwat_od.cover ADD COLUMN fk_distributor  integer;
 ALTER TABLE qwat_od.cover ADD COLUMN fk_status       integer;
 ALTER TABLE qwat_od.cover ADD COLUMN fk_cover_type   integer;
 ALTER TABLE qwat_od.cover ADD COLUMN fk_installation integer;
+ALTER TABLE qwat_od.cover ADD COLUMN fk_precision            integer NOT NULL;
+ALTER TABLE qwat_od.cover ADD COLUMN fk_precisionalti        integer;
+ALTER TABLE qwat_od.cover ADD COLUMN fk_object_reference     integer;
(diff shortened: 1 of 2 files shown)


lbartoletti:
@tudorbarascu can you update your PR for next release, please?

@ponceta @kandre thought?

tudorbarascu:
@lbartoletti I updated the PR. These columns were lost when the covers weren't considered nodes anymore. Now they're back. In the delta file I added a default ``103`` "To be determined" value for ``fk_precision`` and then dropped it afterwards as the fk_precision column should be NOT NULL.

tudorbarascu:
@lbartoletti Can be merged now IMHO. Travis was complaining as I had a ``-`` almost at the end of the delta file and it ignored it.

tudorbarascu:
Is there any reason not to merge this? IMHO it's a regression as we lost these columns.

lbartoletti:
target for release 1.3.3 #250

lbartoletti:
Thanks @tudorbarascu

ponceta:
@tudorbarascu Sorry for latency! Thanks for these!

I've encountered a small bug with the deltas : 

https://github.com/qwat/qwat-data-model/pull/188/files#diff-014985b371786bb83041e7b4f3b80075R15

If you have altitude on existing cover data, these will fail since `fk_precisionalti` and `fk_object_reference` will be NULL.

```
ALTER TABLE qwat_od.cover ADD CONSTRAINT chk_cover_altitude_obj_ref CHECK (fk_object_reference IS NOT NULL OR altitude IS NULL );
ALTER TABLE qwat_od.cover ADD CONSTRAINT chk_cover_altitude_precisionalti CHECK (fk_precisionalti IS NOT NULL OR altitude IS NULL );
```

tudorbarascu:
@ponceta Hmm.. indeed. Then somewhat of an .." update all covers that have altitude and set their fk_obj_ref and fk_precisonalti to TO BE DETERMINED" should be the solution, right?

PR #189, opened 2017-07-18
fill pipe materials table with PE big diameters

tudorbarascu:
delta file will be created later on when the CI will function


tudorbarascu:
@3nids @haubourg Just noticed that the hydrant materials codes are overlapping with pipe materials.

So, https://github.com/qwat/qwat-data-model/blob/master/value_lists/readme is not correct.

PR #190, opened 2017-07-31
Db manager

marioba:
The QWAT db_manager is composed by some parts:
- **Checker** a script that compares 2 databases and shows the differences
- **Dumper** a script to backup and restore a Postgres db
- **Upgrader** a script that apply one or more delta files to an existing database and stores in a table the informations about the applied delta
- **Manager** an interface for managing specific QWAT upgrade procedure


.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: python
 
+notifications:
+  email: false
+
```
m-kuhn: That probably shouldn't go to production

db_manager/dumper.py
```diff
@@ -0,0 +1,106 @@
...
+        command.append('-f')
+        command.append(self.file)
+        command.append(self.__get_dbname())
+
+        try:
+            subprocess.check_output(command, stderr=subprocess.STDOUT)
+
+        except subprocess.CalledProcessError as e:
+            print('pg_dump failed', e.output)
+            raise SystemExit(1)
```
m-kuhn: That sounds a bit hard, I can imagine we will want to run this from inside QGIS in some user interface the future.

system/upgrades_table.sql
```diff
@@ -0,0 +1,17 @@
+
+-- adds informations table used by the new db_manager
```
m-kuhn: "new" will soon no longer be outdated ;)

db_manager/test/test_dumper.py
```diff
@@ -0,0 +1,59 @@
...
+        os.mkdir('/tmp/test_dumper/')
+
+    def test_dump_restore(self):
+        dumper = Dumper(self.pg_service1, '/tmp/test_dumper/dump.sql')
+        dumper.pg_backup()
+
+        dumper = Dumper(self.pg_service2, '/tmp/test_dumper/dump.sql')
+        dumper.pg_restore()
+
+        # postgres > 9.4
```
ponceta: is postgres strictly > 9.4 or >= 9.4 for this?

ponceta:
Does this has some prerequisites? submodules or external librairies?

What are the specs? Postgres and postgis required and tested versions?

.travis.yml
m-kuhn: @marioba can you undo this change?

system/upgrades_table.sql
m-kuhn: and db_manager should be pum ;)

db_manager/test/test_dumper.py
marioba: It should be >=9.4 see [here](https://www.postgresql.org/docs/9.6/static/release-9-4.html). 
I used this only in the unit test modules. In all other places, the sql code is >=9.3 as defined in qwat requirements.

.travis.yml
marioba: @m-kuhn done!

system/upgrades_table.sql
marioba: Fixed!

db_manager/test/test_dumper.py
marioba: It should be >=9.4 see [here](https://www.postgresql.org/docs/9.6/static/release-9-4.html).
I used this only in the unit test modules. In all other places, the sql code is >=9.3 as defined in qwat requirements.

m-kuhn:
Thanks @marioba can you fix the merge conflicts / squash everything?

Next steps:

1. merge this PR
2. integrate view logic
3. fix the deployment of demodata

PR #191, opened 2017-08-03
add automatic valve actuation to fit automatic air release valve

tudorbarascu:


3nids:
@haubourg  shall we consider this as DB model version change?
I have no clear opinion here.

haubourg:
IMHO any change in the structure of the main tables and reference value list is a model change. 
Yet this is very minor change so we could embed that in 1.2.7. 
We should then amend the 1.2.7 update patch here : /update/delta/delta_1.2.7_auto_printmapview.sql
and rename it to delta_1.2.7.sql

How does that sound to you?

3nids:
fine, but let's merge it with 1.2.8 rather than editing an already merged delta.

haubourg:
Ok for a 1.2.8. I proposed that since we didn't released yet the 1.2.7 and too much versions seem to freak out users. 

@tudorbarascu could you start a update patch for 1.2.8 version ?

3nids:
@haubourg this is 1.2.8 https://github.com/qwat/qwat-data-model/pull/194

tudorbarascu:
@haubourg Sure. I added 1.2.9 as @3nids has a pull with 1.2.8

3nids:
merged with https://github.com/qwat/qwat-data-model/pull/194

PR #194, opened 2017-09-13
add hardcoded _schema_visible field to pipes and valves for enhanced …

3nids:
…perfomance on QGIS

This creates a field _schema_visible which determines if
an element is visible in schematic view.
It is calculated by COALESCE(schema_force_visible, value_list.schema_visible).
This allows direct use of this field in QGIS for rule-based labeling and
fixes performance issues if calculation is done on client side.

* schema_visible field is dropped on network_element as it is not used (was introduced for valves at the origin)
* _schema_visible and schema_force_visible are added to pipe and valve tables by the qwat_sys.fn_enable_schemavisible function
* old function qwat_sys.fn_enable_schemaview is enhanced and renamed

### DIFF
--- a/update/delta/delta_1.2.8_schema_visible_perf.sql
+++ b/update/delta/delta_1.2.8_schema_visible_perf.sql
@@ -1,9 +1,26 @@
 /*
-	qWat - QGIS Water Module
+This creates a field _schema_visible which determines if
+an element is visible in schematic view.
+It is calculated by COALESCE(schema_force_visible, value_list.schema_visible).
+This allows direct use of this field in QGIS for rule-based labeling and
+fixes performance issues if calculation is done on client side.
 
-	SQL file :: function to add schema view capability to tables
+* schema_visible field is dropped on network_element as it is not used (was introduced for valves at the origin)
+* _schema_visible and schema_force_visible are added to pipe and valve tables by the qwat_sys.fn_enable_schemavisible function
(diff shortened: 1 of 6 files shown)


3nids:
this does not provide very noticeable performance gain.
But this fixes a few names of triggers and brings a useful field in the tables (pipe/valves).
@haubourg @ponceta any concern to merge this?

update/delta/delta_1.2.8_schema_visible_perf.sql
```diff
@@ -0,0 +1,99 @@
...
+				END IF;
+				RETURN NEW;
+			END;
+		  $$
+			LANGUAGE plpgsql;',
+				_table_name,
+				_vl_table,
+				_fk_field);
+
+		EXECUTE format('CREATE TRIGGER tr_%1$I_schema_visible_update
```
haubourg: Hi @3nids  I spent a long time trtying to understand why migration failed here. 
In  system/fn_enable_schemaview.sql, triggers are sill named  I_schema_view_update when they are named I_schema_visible_update in the update.

system/fn_enable_schemaview.sql
```diff
@@ -1,26 +1,60 @@
 /*
```
haubourg: I would rename this file system/fn_enable_schemavisible.sql and modify the init script accordingly to avoid naming confusion. Any opinion?

PR #197, opened 2017-09-25
Pipe protection

tudorbarascu:
For PE and Steel pipes there were only OLD pipes selection but we have heavy traffic pipes that are protected with New pipes (PE, Steel or PVC).
Thoughts?


haubourg:
@ponceta @dsavary @varrieta any opinion on those new protection values?

ponceta:
Any ADDs are welcome IMHO.

haubourg:
thanks !

Issue #195, opened 2017-09-20
1.2.8 update fail and weird deltas


ponceta:
@3nids, @haubourg 1.2.8 Can't be updated automatically.

```
UPDATE 4468
psql:/home/pulmexa/qwat-data-model-original/update/delta/delta_1.2.8_schema_visible_perf.sql:90: ERREUR:  n'a pas pu supprimer table qwat_od.pipe colonne schema_force_visible_old car d'autres objets en dépendent
DETAIL:  vue qwat_od.vw_export_pipe dépend de table qwat_od.pipe colonne schema_force_visible_old
HINT:  Utilisez DROP ... CASCADE pour supprimer aussi les objets dépendants.
```

This can be fixed by removing and recreating the vw_export_pipe as in https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql#L312-L318

I can't understand why https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql is creating some views since update process is rewriting views at https://github.com/qwat/qwat-data-model/blob/master/update/upgrade_db.sh#L178

@sylvainbeo is there a reason why views are not dropped before update? 

If not, I would advise avoiding "create view" or "comment on view" statements in the deltas section and keep these in the view creation scripts.

Last, I get some warnings on 1.2.8 :
`psql:/home/pulmexa/qwat-data-model-original/update/delta/delta_1.2.8_schema_visible_perf.sql:88: ATTENTION:  [qwat_sys.if_modified_func] - Trigger detected NULL hstore. ending`
Can these be ignored?

ponceta:
OK.

If we drop the views, before updating datamodel, this will cause some issues in:

In delta 1.0.1
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.0.1_06062016.sql#L50
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.0.1_06062016.sql#L70

In delta 1.1.0
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql#L273
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql#L279
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql#L285
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql#L291

[...]

PR #198, opened 2017-09-26
Fix 'upgrade table' delta number and update submodule pum

marioba:
- Fix the version number of the delta file that creates the 'upgrade table' to 1.2.9.
- Update the submodule pum. This pum's version returns an exit code in case of error so Travis can understand it correctly.


marioba:
Now Travis checks fail because of https://github.com/qwat/qwat-data-model/issues/195

3nids:
drop_views.sql should be run as a _pre_ script

m-kuhn:
@marioba can you implement the static pre-all.sql? I guess that should be simple

PR #201, opened 2017-10-16
fixup for 1.2.8

haubourg:
So, delta file for 1.2.8 is fixed, as is upgrade_db.sh script for local updates. it works for migration from 1.2.6 to 1.2.8 at least. 
I couldn't find any working configuration to fix the travis script for all possible use cases because of dependencies to generated view in manual views. 

So, as long as this conformity checking script has a few days to live, I will stop wasting time on that old script, unless someone finds a nice solution to that. 

I'll squash commit history before merge.


sylvainbeo:
Seems ok for me. 
I also think this is the best solution considering the upcoming PUM.

m-kuhn:
Isn't PUM already in place?

haubourg:
@m-kuhn Well, old migration control scripts are still there in Travis at least.

m-kuhn:
Is there a good reason for this or are we just waiting for the official decision tomorrow to shut things down?

haubourg:
My goal wa to end up with a clean migration process until 1.2.9, because some QWAT are in production and rely on that migration process.
Since we fail in fixing that old process, I see no objection in shutting this down. @sylvainbeo can you do that.?

m-kuhn:
Wouldn't it be even easier to migrate these using pum as well?

haubourg:
I hope so, but the QWAT project generation process is full of surprises. Not having the double conformity check will anyway make it a lot easier

PR #203, opened 2017-10-18
fix 2 for 1.2.8

haubourg:
This time, 
changing trigger naming in init process and restore correct naming in delta file. 
Also Rename fn_enable_schemaview.sql to fn_enable_schemavisible.sql and fix init_qwat.sh to use this new name. This will keep naming convention correct. 
The migration works from 1.2.6 on my install up to 1.2.9 using the fixed upgred_db.sh. 

@3nids I'm letting you merge or fix it using another method.


3nids:
this is fine with me, but it comes back to its previous state apart from the file name, no?

haubourg:
Nope, previous state we had an inconsistency between trigger names generated by the delta and trigger names generated by the init process. We're now inline.

3nids:
perfect then :)

PR #205, opened 2017-10-25
Pum integration2

sylvainbeo:
@m-kuhn @haubourg Hi, have a look to this PR. I've added pre and post-all delta file + a pre for 1.2.8. 

Please comment.


3nids:
is that a green light ????
haven't seen one for quite a while around here...
good job to all of you!!!

m-kuhn:
Great stuff, that's good to see.

@marioba any feedback from your side concerning the use of `os.system`? Alternatively there would be the `subprocess` module which could also intercept `stdout` and `stderr`, should that be sent back to Pum, and of course `psycopg`. Not sure how you envisioned this.

marioba:
👍for the green light...
 
In the `DeltaPy` abstract class, there is the `write_message` method designed to write the subclass outputs. Currently, the method doesn't make much more than a simple `print()` but I think it would be good to have a centralized output so, if we want to manage it differently in the future, it's easy to modify it. 
And maybe it would be also a good idea to define some exception types that the subclasses can raise in the `run` method...

m-kuhn:
Thinking about this again, it might be nice to send all sql through psycopg2. At least from a transaction perspective it would be nice to send everything through the same interface and have total control to commit or rollback.

sylvainbeo:
Ok all, we wait for @haubourg  to have a look, and we merge.

ponceta:
Thanks for the good work, this will be great to have this!

There's a small error here :

https://travis-ci.org/qwat/qwat-data-model/builds/293131379#L2870

Probably linked to SRID not beeing initialized somehow?

Il like the green tick but having errors and still been green can be a false positive for me.

The auth token is not working too?

sylvainbeo:
@ponceta Ah oups. did not see this one, because travis was collapsing the lines. Yes indeed, it is because of the missing SRID. I have a look.

m-kuhn:
Can someone check if the `GH_TOKEN` has the `public_repo` access set?

m-kuhn:
@sylvainbeo I guess you need to check the return value of the `psql` call.

sylvainbeo:
@m-kuhn The problem is on the post-all process. No sql is launched directly, just .sh 
I have a look.

sylvainbeo:
SRID problem seems to be solved. 

Now i see another one  'schema "qwat_od" does not exist'

sylvainbeo:
+ @m-kuhn 

> Traceback (most recent call last):
  File "./.deploy/create_release.py", line 154, in <module>
    main()
  File "./.deploy/create_release.py", line 144, in main
    url='{release_url}?name={filename}'.format(release_url=release['upload_url'][:-13], filename=filename)
KeyError: 'upload_url'

m-kuhn:
@sylvainbeo , that normally happens when there's something broken with the api token

> Can someone check if the GH_TOKEN has the public_repo permission set?

m-kuhn:
That should help: https://github.com/qwat/qwat-data-model/pull/206

sylvainbeo:
I'm not a GH_TOKEN expert. here is the configuration for the tken (qwat_admin user)

image

m-kuhn:
Looks good I think

m-kuhn:
Ah wait... Actually deploy should only happen on tags...

sylvainbeo:
@m-kuhn Indeed no tag.

sylvainbeo:
Anyone has an idea on " schema "qwat_od" does not exist"..very weird.

sylvainbeo:
mmmm... i have one. Maybe the .sh scripts are not launched on the right DATABASE.

sylvainbeo:
Fixed !

sylvainbeo:
Question: in the post-all.py file, i specify qwat_test as the output DB... maybe this should be a param.

haubourg:
Merging that at last!

m-kuhn:
:tada:

PR #209, opened 2017-11-29
better upgrade_db.sh

haubourg:
more verbose, better usage options and help. 
Also contains first attempts for adding local directory containing delta file


haubourg:
merging

PR #210, opened 2017-11-30
1.3.1 version - fixes altitude sync for valves (was old 1.2.9 dev branch)

haubourg:


haubourg:
@3nids does that look good to you?

PR #212, opened 2017-12-04
qwat_prod database hardcode replaced with pgservice argument

nliaudat:


haubourg:
Good catch, thanks!

nliaudat:
great pleasure ;)

PR #213, opened 2017-12-05
use a full demo data dump instead of empty Db

haubourg:
using 1.2.1 full dump instead of 1.0.0. This will allow the automatic demo data upgrade. 
We don't have many users in production between 1.0.0 and 1.2.1, so that shouldn't be a big issue not to test each time updates from 1.0.0


haubourg:
nah, just wasn't using the right url for github download. 
Now I have an issue with the dump that doesn't include the postgis create extension clause. 
let's re-dump a new one.

PR #216, opened 2017-12-26
add missing SIRE column for remotes

3nids:
specific to extension ch_vd_sire

an update will come later for the extension.
but up to now, there was a dedicated column (code_sire) in many VL tables.

this could be fixed later (by having a proper extension management).

I'd like to get this merged as this just fix things according to current approach.


3nids:
objections to merge?

ponceta:
Seems fine to me. Maybe assign a reviewer could be a good practice @vpicavet ?

haubourg:
@3nids We need to update the system/CURRENT_VERSION.txt to 1.3.2 if I get it right. Can you check that?

3nids:
@haubourg done

PR #218, opened 2017-12-27
remove search view

3nids:
this is an oldie, and was specific to the SIGE

I think this doesn't need to have a delta since it's only view related.
It can be merged and wait for next release to be applied.


ponceta:
Nice cleaning. Thanks!

PR #220, opened 2018-02-06
Do not use --section data when creating data+structure dumps

elemoine:
After a sucessfull build the Travis process executes `.deploy/create_release.py`. And, when TRAVIS_TAG is set, the Python script creates database dumps and publishes them as release assets on GitHub. `create_release.py` is supposed to create both data-only and data-and-structure dumps. But in fact it creates data-only dumps only.

This PR fixes this by removing `--section data` for the data-and-structure dumps. I tested it locally by executing `create_release.create_dumps` in a Python console. And I was able to pg_restore the data-and-structure dump into an empty qwat database.


haubourg:
thanks @elemoine !

elemoine:
@haubourg should we update the structure+dumps builds on the release page?

haubourg:
On it!

haubourg:
@3nids @elemoine @sylvainbeo , I'm unsure about how to retrigger a travis build for tag 1.3.1 and then have new release generated , with fixed dumps inside.
We now have new commits in master on the data model, with some pull requests pending that would be a great candidate for 1.3.2 . So dropping tag 1.3.1 and recreating it is not a good practice now. But tagging 1.3.2 might be prematurate. 
Any opinion on how to fix the dump in the past releases ?

elemoine:
For 1.3.1 you can just manually generate the struct+data dumps and manually update the dumps on the release page. I wouldn't change the 1.3.1 Git tag, and attempt to trigger Travis for this.

sylvainbeo:
Agree with @elemoine . Let's do it manually.

PR #222, opened 2018-02-12
Standardize shebangs

lbartoletti:
Like https://github.com/qgis/QGIS/pull/6229

Todo:

[x] Tests python2/python3?
[x] Clean old python stuff?
[x] Pep8

### DIFF
--- a/ordinary_data/functions/rewrite_functions.sh
+++ b/ordinary_data/functions/rewrite_functions.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
--- a/ordinary_data/views/export/export_hydrant.py
+++ b/ordinary_data/views/export/export_hydrant.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function
 from __future__ import absolute_import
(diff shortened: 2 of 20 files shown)


haubourg:
If I get it well, your changes will require python3 by default ? It wasn't the case yet, except for PUM that recently introduced Python 3 dependency

lbartoletti:
Yes.
Some scripts are already with python3, travis run with python 3.5, so I thought there was a switch in progress.
Since python3 is needed for PUM, potentially this can reduce dependencies to python2 (and remove from __future__ from the scripts)

Otherwise, I modify the scripts but with python2 instead of python3
In all cases, in order to be really portable, you must specify the version of python used (/usr/bin/env python does not necessarily exist and can point to python2 or python3 depending on the linux distributions or other OS)

haubourg:
ok good!

@sylvainbeo @3nids @elemoine @ponceta Do you agree for a complete and explicit switch to python3 in QWAT ? 
We'll need to add this in the README and doc too.

ponceta:
This could be documented [here](http://qwat.readthedocs.io/en/latest/requirements.html#server-side) too, as the python version was never defined and python2 was default.

Prerequisites should be set as link to the doc in the [README.rst](https://github.com/qwat/QWAT/blob/master/README.rst) since there is an install section.

By the way, PostgreSQL (> 9.3) and PostGIS, the spatial extension (> 2.1) are historic prerequisites for qwat but tests are made with PostgreSQL 9.6 and Postgis 2.3.

sylvainbeo:
Late, but yes. Source code is fully compatible.

PR #225, opened 2018-02-19
Fixes invalid char in system/audit_views.sql

lbartoletti:
More explicit:
```
-<U+FEFF>/* Audit views only for those having explicit triggers handling data editing*/
+/* Audit views only for those having explicit triggers handling data editing*/
```


haubourg:
@sbe @elemoine @lbartoletti here is the kind of change tha puzzles me according to versioning guidelines. 
technically, are we sure that the function is updated on each PUM migration? At what point should we explicitly version the database (using current_version and internal PUM metadata ) ? triggers, views, and stored procedures are changed here, but we don't release explicitly a version. That grey are needs to be explicitly documented to me

PR #226, opened 2018-02-20
Numversion init qwat

lbartoletti:
Don't hardcode version in init_qwat.sh

### DIFF
--- a/init_qwat.sh
+++ b/init_qwat.sh
@@ -36,6 +36,7 @@ fi
 
 eval set -- "$ARGS";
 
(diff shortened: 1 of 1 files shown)


elemoine:
I personally like good Git commit messages and clean Git history. Would you mind squashing the commits into one commit and provide a more explicit commit message like "Do no hardcode NUMVERSION in init_qwat.sh".

haubourg:
NIce, I don't know why the harcoded version was still here, we created the current_version file for that purpose. Thanks for taking care of that!

PR #227, opened 2018-02-20
New delta for deleting old views not required anymore

sylvainbeo:


update/delta/delta_1.3.3_remove_old_views_and_triggers.sql
```diff
@@ -0,0 +1,35 @@
+
```
elemoine: Why is this a 1.3.3 delta, as opposed to 1.3.2? As far as I understand the current release is 1.3.1, so why not extending the current delta_1.3.2 file instead of creating a new one? This question is mainly for me to undertand…

update/delta/delta_1.3.3_remove_old_views_and_triggers.sql
```diff
@@ -0,0 +1,35 @@
...
+-- From earlier versions of QWAT there was some qwat_od.vw_installation_xxx views, which are not required anymore.
+
+DROP VIEW IF EXISTS qwat_od.vw_installation_pressurecontrol;
+DROP VIEW IF EXISTS qwat_od.vw_installation_pump;
+DROP VIEW IF EXISTS qwat_od.vw_installation_source;
+DROP VIEW IF EXISTS qwat_od.vw_installation_tank;
+DROP VIEW IF EXISTS qwat_od.vw_installation_treatment;
+DROP VIEW IF EXISTS qwat_od.vw_installation_chamber;
+
+DROP function IF EXISTS qwat_od.ft_installation_chamber_delete();
```
elemoine: s/function/FUNCTION for consistency?

update/delta/delta_1.3.3_remove_old_views_and_triggers.sql
sylvainbeo: @elemoine Good point. 

1/ I think that if there is a 1.3.2 delta, so the version should be 1.3.2. So there's an error here. 
2/ Why not merging the DLETE in the 1.3.2, but the delta name will have a wrong name
3/ Agree for the FUNCTION, corrected.

update/delta/delta_1.3.3_remove_old_views_and_triggers.sql
sylvainbeo: Corrected

update/delta/delta_1.3.3_remove_old_views_and_triggers.sql
elemoine: > 1/ I think that if there is a 1.3.2 delta, so the version should be 1.3.2. So there's an error here.

Yep, maybe. I am not familiar with the qWat release process. So I am not sure.



> 2/ Why not merging the DLETE in the 1.3.2, but the delta name will have a wrong name

1.3.2 is not released yet, so I tend to think that you can freely change the delta name. (I'd love that someone confirm this.)

Also, in my opinion, delta names don't make too much sense. If it was me I'd just numbers. For example: delta_1.3.2_1.sql, delta_1.3.2_2.sql, etc. For the record, I think Pum does support multiple delta files per version already, but I haven't verified it. I am not suggesting to switch from delta names to delta numbers right now, but maybe this is something that could be discussed at some point.

update/delta/delta_1.3.2_fix_valve_missing_constraints.sql
```diff
@@ -1,7 +0,0 @@
-
-
--- Missing valve constraints
-
--- ALTER TABLE qwat_od.valve ADD CONSTRAINT chk_valve_altitude_obj_ref CHECK (fk_object_reference IS NOT NULL OR altitude IS NULL );
--- ALTER TABLE qwat_od.valve ADD CONSTRAINT chk_valve_altitude_precisionalti CHECK (fk_precisionalti IS NOT NULL OR altitude IS NULL );
-SELECT 1;
```
elemoine: This was a rather strange delta file! Do you have explanations for this?

update/delta/delta_1.3.2_various_fix.sql
```diff
@@ -0,0 +1,40 @@
+
+
+-- Missing valve constraints
+
+-- ALTER TABLE qwat_od.valve ADD CONSTRAINT chk_valve_altitude_obj_ref CHECK (fk_object_reference IS NOT NULL OR altitude IS NULL );
+-- ALTER TABLE qwat_od.valve ADD CONSTRAINT chk_valve_altitude_precisionalti CHECK (fk_precisionalti IS NOT NULL OR altitude IS NULL );
```
elemoine: Do we want to keep that? For what reasons would we?

update/delta/delta_1.3.2_fix_valve_missing_constraints.sql
sylvainbeo: That was in that PR  https://github.com/qwat/qwat-data-model/pull/219/files , but i don't remember why it is commented... I propose to leave the comment for the moment, and clean them once we'll be sure it's not an error.

elemoine: You are the one who commented it in https://github.com/qwat/qwat-data-model/commit/d3dbec4b6f39e7789947e40e1b649690d9959f61.

sylvainbeo: Yes. Don't remember the reason.

elemoine: In that case I don't see the reason for keeping them :)

sylvainbeo: Ok right. Deleted.

haubourg:
Hi guys, good cleanup. 
Every database change should be versionned. So we need a change for "current_version" file and for the PUM metadata. 
We proposed in the general meeting last tuesday a semver rule about this. 
In that case, a minor fix, that would be smething like "1.3.2-01"
@sylvainbeo can you change to that behavior?

sylvainbeo:
Done. Version set to 1.3.2-01

PR #228, opened 2018-02-22
Fixes init_qwat and upgrade_db.sh

lbartoletti:
1. Replace more gnued/linuxism script
1.1. getopt
Using long args with getopt is a linuxism. So I include a "compat mode" for it. More details on [stackoverflow](https://stackoverflow.com/a/7948533/8299570)
1.2. sed
`sed r` is the GNU sed version. Use coreutils with gsed for non linux. Or easiest, use `cat`
1.3. sleep
Use only the number of seconds without the suffix.

2. upgrade_db.sh
2.1. SCRIPTDIR
Add the location of the script. Since you can run it everywhere
2.2. DB info from pg_service.conf
TODO is done. Get the host and the port for dbcomp and dbtest.
Maybe add an option for the location of .pg_service.conf?
Wouldn't it be better if this part were done by PUM?

### DIFF
--- a/update/upgrade_db.sh
+++ b/update/upgrade_db.sh
@@ -76,7 +76,7 @@ if [ ! -f $VERSION_FILE ]; then
     echo
     exit 0
 fi
-VERSION=$(sed 'r' $VERSION_FILE)
+VERSION=$(cat $VERSION_FILE)
 
 # set -- "${POSITIONAL[@]}" # restore positional parameters
 echo "Parameters:"
@@ -94,7 +94,7 @@ echo
 if [[ $CLEAN -eq 1 ]]; then
     printf "\n${BLUE}Cleaning  Option set:${NC}\n\n"
 
-    sleep 1s
+    sleep 1
 
     # Read DB info from pg_service.conf file
     DBCOMP_NAME=$(sed -n -e "/^\[qwat_comp]/,/^\[/ p" ~/.pg_service.conf | grep "^dbname" | cut -d"=" -f2)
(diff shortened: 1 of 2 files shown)


sylvainbeo:
Looks good to me.

haubourg:
reviewed also by @mhugo :) >> merging

update/upgrade_db.sh
```diff
@@ -94,43 +95,59 @@ echo
...
     DBCOMP_USER=$(sed -n -e "/^\[qwat_comp]/,/^\[/ p" ~/.pg_service.conf | grep "^user" | cut -d"=" -f2)
+    DBCOMP_HOST=$(sed -n -e "/^\[qwat_test]/,/^\[/ p" ~/.pg_service.conf | grep "^host" | cut -d"=" -f2)
+    DBCOMP_PORT=$(sed -n -e "/^\[qwat_test]/,/^\[/ p" ~/.pg_service.conf | grep "^port" | cut -d"=" -f2)
     DBTEST_NAME=$(sed -n -e "/^\[qwat_test]/,/^\[/ p" ~/.pg_service.conf | grep "^dbname" | cut -d"=" -f2)
     DBTEST_USER=$(sed -n -e "/^\[qwat_test]/,/^\[/ p" ~/.pg_service.conf | grep "^user" | cut -d"=" -f2)
-    # TODO get also password, port and host
+    DBTEST_HOST=$(sed -n -e "/^\[qwat_test]/,/^\[/ p" ~/.pg_service.conf | grep "^host" | cut -d"=" -f2)
+    DBTEST_PORT=$(sed -n -e "/^\[qwat_test]/,/^\[/ p" ~/.pg_service.conf | grep "^port" | cut -d"=" -f2)
+
+    if [ ! -z "$DBCOMP_HOST" ]; then
```
elemoine: Any difference between `! -z` and `-n`?

PR #229, opened 2018-02-22
Missing spatial indexes for some qwat_od's tables

lbartoletti:
Missing spatial indexes for some qwat_od's tables:
* cover - geometry
* cover - geometry_polygon
* crossing - geometry
* leak - geometry
* surveypoint - geometry
* valve - handle_geometry

### DIFF
--- a/ordinary_data/pipe/od_crossing.sql
+++ b/ordinary_data/pipe/od_crossing.sql
@@ -21,6 +21,7 @@ WITH (
 );
 
 ALTER TABLE qwat_od.crossing ADD COLUMN geometry geometry('POINT',:SRID) NOT NULL;
+CREATE INDEX crossing_geoidx ON qwat_od.crossing USING GIST ( geometry );
 
 CREATE OR REPLACE FUNCTION qwat_od.ft_controled_crossing()
 RETURNS trigger AS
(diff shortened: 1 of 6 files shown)


elemoine:
Just for me to understand: are those indexes required for good performance in QGIS?

ordinary_data/pipe/od_crossing.sql
```diff
@@ -20,6 +20,8 @@ WITH (
   OIDS=FALSE
 );
 
+CREATE INDEX crossing_geoidx ON qwat_od.crossing USING GIST ( geometry );
```
elemoine: Here the index is created before the creation of the geometry column.

update/delta/delta_1.3.2_001_missing_spatial_indexes.sql
```diff
@@ -0,0 +1,9 @@
+-- Missing spatial indexes
```
elemoine: Should we agree on the delta file name scheme before actually using it? @haubourg any opinion about this?

lbartoletti:
> Just for me to understand: are those indexes required for good performance in QGIS?
Yes, especially leak, cover and surveypoint can be used frequently.

haubourg:
@3nids Looks good to me. Any objections to merge?

PR #232, opened 2018-02-26
Homogenize naming of 1.3.2 deltas

elemoine:
Following the proposal made in https://github.com/qwat/QWAT/issues/259 we now use multiple delta files pers version. But we are currently using different naming schemes in the repo. With this PR I am suggesting to use a three-digit prefix, from `000` to `999`. In this way we can order delta files independently from the meaningful part of their names.


sylvainbeo:
Good. Pum has been updated to follow that rule  ?

elemoine:
Yes, Pum was updated. And a new release was published on pypi.

PR #233, opened 2018-02-26
Deactivate tables auditing during migrations

elemoine:
Fixes https://github.com/qwat/QWAT/issues/248.

Depends on #232, which should be merged first.

### DIFF
--- a/update/delta/pre-all.sql
+++ b/update/delta/pre-all.sql
@@ -1,0 +1,14 @@
+DO $$
+BEGIN
+
+    -- before 1.3.2 deactivate_audit_tables did not exist, so we test that the function
+    -- exists prior to calling it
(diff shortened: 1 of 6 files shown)


elemoine:
Now rebased and ready for review.

update/delta/pre-all.sql
```diff
@@ -0,0 +1,14 @@
+DO $$
+BEGIN
+
+    -- before 1.2.3 deactivate_audit_tables did not exist, so we test that the function
```
lbartoletti: 1.3.2?

elemoine: Good catch! Fixed.

haubourg:
Merci !

PR #235, opened 2018-02-27
Update requirements.txt: psycopg2->psycopg2-binary

lbartoletti:
The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead.


elemoine:
I was about to do the same. Please merge when Travis is happy.

PR #236, opened 2018-02-27
Make the Travis build stop on pum errors

elemoine:
This WIP at the moment.

Currently the Travis will continue when one of the steps fails in the `script` part. Instead we want the Travis build to stop as soon as something fails. So this PR introduces a `travis_script.sh` script to fix the problem. I also created `travis_before_script.sh` for consistency reasons and to make the code clearer.


elemoine:
This works in Travis. Now ready for review.

Issue #199, opened 2017-09-26
Hardcoded SRS in deltas


ponceta:
SRS should not be hardcoded :

delta 1.1.0
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql#L48-L50
https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.1.0.sql#L296

3nids:
@marioba does pum handle parameter to psql commands?
such as psql -v SRID=2056

marioba:
@3nids No, not at the moment. Pum only executes the code found in the sql files as it is using psycopg2.

elemoine:
The SRS is hardcoded in Pum Python scripts as well. See https://github.com/qwat/qwat-data-model/blob/master/update/delta/post-all.py#L14 for example.

3nids:
I just added variables for SQL deltas, it shouldn't be a big deal to add them for python deltas. Let me know if you want me to add them.

ponceta:
@kandre in what version of QWAT are you? 

Please reopen this if you encounter any troubles when migrating.

kandre:
@ponceta FYI we have 1.2.8 in production but not sure if I will be concerned by this issue or not. I'll see that when upgrading thanks !

3nids:
isn't this a high priority?

ponceta:
https://github.com/qwat/qwat-data-model/blob/master/update/delta/post-all.py#L14

@3nids I missed the post-all .py delta file. This is indeed a high priority bug.

Actually I can't update anymore since postgres 9.5 or higher is required so I'm stuck in 1.3.2

lbartoletti:
Closed with a workaround in https://github.com/qwat/qwat-data-model/pull/298
I will check the bug with pum

PR #239, opened 2018-03-02
Fixes views without SRID and not populated

lbartoletti:
Using a cast for vw_pipe_schema
Using a cast for export_pipe with a workaround in python


sylvainbeo:
Looks ok for me. Seeing the 21781 in the sql file, it reminds me that we could think about a way to externalize it.

lbartoletti:
Yes. I didn't understand at first why it was harcoded.

haubourg:
this is #199 
And yes it is important to externalise it

update/delta/delta_1.3.2_007_fixes_views.sql
```diff
@@ -0,0 +1,256 @@
+-- View: qwat_od.vw_export_pipe
+
+-- DROP VIEW qwat_od.vw_export_pipe;
+
+CREATE OR REPLACE VIEW qwat_od.vw_export_pipe AS
```
elemoine: I need to check but I am not sure we need deltas for views.

haubourg:
Correct. no Delta needed for views if they are created via ordinary_data/views scripts

lbartoletti:
Removed. Thanks.

haubourg:
Thanks Loïc, merging!

PR #243, opened 2018-03-09
Make Travis create structure-only dumps

elemoine:
This modifies the `create_release.py` script to also create structure-only dumps. Structure-only dumps are useful when one wants to get an empty QWAT database without having to clone the Git repo and run `init_qwat.sh`.

As discussed @lbartoletti will take care of updating the docs to take this setup path into account.


elemoine:
PR updated to also include value list data only dumps.

haubourg:
LGTM

PR #246, opened 2018-03-23
Fix, improve and test the sire extension's initialization

elemoine:
This PR fixes the ch_vd_sire extension related issues reported in https://github.com/qwat/qwat-data-model/pull/217#issuecomment-374695096.

It also sets up a delta dir specific to the ch_vd_sire extension, providing pre-all.py and post-all.py scripts for dropping and recrating the ch_vd_sire views.

This PR also modfies the travis_script.sh script to start testing the ch_vd_sire extension on Travis. Adding this revealed bugs in QWAT's pre-all.py and post-all.py scripts, which this PR fixes as well.

Please review.


ordinary_data/views/drop_views.sql
```diff
@@ -1,53 +1,53 @@
 
 /* schematic */
-drop view if exists qwat_od.vw_pipe_schema_error;
-drop view if exists qwat_od.vw_pipe_schema;
-drop view if exists qwat_od.vw_pipe_schema_merged;
-drop view if exists qwat_od.vw_pipe_schema_visibleitems;
-drop view if exists qwat_od.vw_pressurezone_simplified;
+drop view if exists qwat_od.vw_pipe_schema_error cascade;
```
3nids: Please don't write `CASCADE`, there's good chance to erase views from the user. To me it's hard requirement that we never write cascade in a delta or in the view scripts.

extensions/ch_vd_sire/delta/pre-all.py
```diff
@@ -0,0 +1,15 @@
+import os
+from pum.core.deltapy import DeltaPy
+
+
+class DropViews(DeltaPy):
```
3nids: I like a lot the idea of setting drop/rewrite views like this, it's cleaner and more consistent.

Does PUM handle this already? 
Also, is it done in proper order (extensions should be run first on `pre` and and last on `post`)?

elemoine: We extended Pum so it can work with multiple delta directories, for the purpose of QWAT extensions effectively.

Currently, the execution order of `pre` and `post` is the same. We can easily change that in Pum, and have the execution of `post` being the reverse of `pre`. I'll have a closer look.

ordinary_data/views/drop_views.sql
elemoine: You're right. Let's see how we can do otherwise.

elemoine:
Thanks @3nids for your good comments!

I removed the commit that added `CASCADE` to the `DROP VIEW` statements. And I changed Pum (in  my own [branch](https://github.com/opengisch/pum/compare/master...elemoine:ele_delta-dir) for now) so that the `pre` scripts of the extension are executed before the `pre` scripts of QWAT core. Things work as expected.

So I'll submit a PR to Pum, and remove commit 288f087 when the Pum PR is merged.

elemoine:
See https://github.com/opengisch/pum/pull/30 for the PR to Pum.

marioba:
Released in pum [0.5.13 ](https://github.com/opengisch/pum/releases/tag/0.5.13)

extensions/ch_vd_sire/delta/pre-all.py
3nids: yes, I think this is mandatory.

extensions/ch_vd_sire/init_sire.sh
```diff
@@ -0,0 +1,90 @@
...
+
+if [[ -z $PGSERVICE ]]
+then
+  echo "Error: no PG service provided; either use -p or set the PGSERVICE environment variable."
+  exit 1
+fi
+
+if [[ "$DROPSCHEMA" -eq 1 ]]; then
+	psql service=${PGSERVICE} -v ON_ERROR_STOP=1 \
+         -c "DROP SCHEMA IF EXISTS qwat_ch_vd_sire CASCADE"
```
3nids: Do we really need a cascade here?

3nids:
This sounds very sensible to me, good job!

extensions/ch_vd_sire/init_sire.sh
elemoine: I just used the same approach as in `init_qwat.sh`. It makes sense to be consistent IMO.

3nids: ok, makes sense!

.build/travis_script.sh
```diff
@@ -65,4 +65,15 @@ echo "travis_fold:end:test-and-upgrade"
 # Run a last check between qwat_prod and qwat_comp
 pum check -p1 qwat_prod -p2 qwat_comp -i views rules
 
+# Extend qwat_prod with ch_vd_sire
+printf "travis_fold:start:init-sire\nExtend database with ch_vd_sire"
+$TRAVIS_BUILD_DIR/extensions/ch_vd_sire/init_sire.sh -p qwat_prod -s 21481
```
lbartoletti: 21781

elemoine: Fixed, thanks a lot.

elemoine:
Thank you @lbartoletti and @3nids for the reviews.

PR #249, opened 2018-04-09
WIP: Make changes for Pum 0.6

elemoine:
In Pum 0.6 (and higher) the version flag was changed for `-v` to `-V`. This PR accomodates for this change.

Also, there's a regression in Pum 0.6. See https://github.com/opengisch/pum/pull/38. Commit f3e233c tests that the fix for the regression does work for QWAT. I'll remove that commit when there's a new Pum release on pypi.

### DIFF
--- a/init_qwat.sh
+++ b/init_qwat.sh
@@ -97,7 +97,7 @@ done
 
 
 # Check if pum is installed (will fail and stop the script in case of error)
-PUM_VERSION=$(pum -v)
+PUM_VERSION=$(pum -V)
 echo "PUM version: $PUM_VERSION"
 echo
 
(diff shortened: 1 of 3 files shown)


elemoine:
Ok, https://github.com/opengisch/pum/pull/38 fixes the problems. Let's wait for a new Pum release before merging this.

3nids:
@elemoine would you prefer me to revert the -v/-V change or we keep the new one?

elemoine:
> @elemoine would you prefer me to revert the -v/-V change or we keep the new one?

Yes, that would make sense to me. I think `-v` is more natural.

lbartoletti:
@elemoine Do you think this can be fixed and integrated into 1.3.3?

elemoine:
I've just checked with Pum 0.7.1 (the latest version), and `pum -v` (lower-case `v`) does output the version again. So this PR can just be closed.

PR #252, opened 2018-04-09
Fix valve_count in pipe

lbartoletti:
Fix https://github.com/qwat/QWAT/issues/244


3nids:
@haubourg not in this case. This are specific triggers (règle métier) which are not automatically generated.

ordinary_data/valve/tr_valve_pipe.sql
```diff
@@ -5,7 +5,49 @@
 */
 
 /* UPDATES NUMBER OF VALVES AND CLOSED/OPEN FOR PIPES */
-CREATE OR REPLACE FUNCTION qwat_od.fn_pipe_update_valve(_pipe_id integer) RETURNS VOID AS
+CREATE OR REPLACE FUNCTION qwat_od.fn_pipe_delete_valve(_valve_id integer)
+  RETURNS void AS
+$BODY$
+BEGIN
+
+IF (SELECT _valve_count > 1 FROM qwat_od.pipe WHERE id = (
```
3nids: would it do any harm to skip this check?
if the field is not correct, it might be good to go further.
And I don't think this is too costless, is it?

ordinary_data/valve/tr_valve_pipe.sql
```diff
@@ -5,7 +5,49 @@
...
+					SELECT fk_pipe
+					FROM qwat_od.valve
+					WHERE id = _valve_id))
+THEN
+			UPDATE qwat_od.pipe
+			SET
+				_valve_count  = valve_group.vcount,
+				_valve_closed = valve_group.vclosed
```
3nids: can you detail the logic?
it's not possible to handle this in an AFTER trigger?

ordinary_data/valve/tr_valve_pipe.sql
```diff
@@ -67,23 +114,35 @@ COMMENT ON TRIGGER tr_valve_pipe_delete ON qwat_od.pipe IS 'Trigger: when deleti
...
-		PERFORM qwat_od.fn_pipe_update_valve(NEW.id);
+		IF TG_OP = 'INSERT' THEN
+			PERFORM qwat_od.fn_pipe_update_valve(NEW.id);
+		END IF;
+		IF TG_OP = 'DELETE' THEN
+			PERFORM qwat_od.fn_pipe_delete_valve(OLD.id);
+		END IF;
+		IF TG_OP = 'UPDATE' THEN
+			PERFORM qwat_od.fn_pipe_delete_valve(OLD.id);
+			PERFORM qwat_od.fn_pipe_update_valve(NEW.id);
```
3nids: yep, if you can give details why we need two steps here.

3nids:
Would you have time to write a test for this?
I know it's not the funny part, but I think the situation should be improved there.

ordinary_data/valve/tr_valve_pipe.sql
lbartoletti: For one case: when you change the fk_pipe id for a valve

ordinary_data/valve/tr_valve_pipe.sql
lbartoletti: It's the same as update instead of I excude the valve_id from the count (for bool_.

3nids: yep, I got the idea, but it seems a bit fragile to me (there is no control that the -1 actually corresponds to anything)
That's why I'm asking if we can't have a simpler function in an after trigger (hence no need for the -1).

lbartoletti: Done

haubourg:
> @haubourg not in this case. This are specific triggers (règle métier) which are not automatically generated.

OK, maybe we could a clear metadata tag for autogenerated triggers and hand maid's one ? 
@lbartoletti opinion?

ordinary_data/valve/tr_valve_pipe.sql
```diff
@@ -67,23 +119,49 @@ COMMENT ON TRIGGER tr_valve_pipe_delete ON qwat_od.pipe IS 'Trigger: when deleti
...
+		
+		RETURN OLD;
+	END;
+$BODY$
+LANGUAGE plpgsql;
+
+COMMENT ON FUNCTION qwat_od.ft_before_valve() IS 'Trigger: when deleting a valve, reevaluate pipes for number of valves on them.';
+
+CREATE TRIGGER tr_valve_delete_trigger
+  BEFORE DELETE
```
3nids: so, my question was mainly why don't you do an AFTER DELETE here?

3nids:
Another thing, is the moving of valve handled (I think it used to be).
Could you add a move to the test too (move off pipe and move to pipe)?

ordinary_data/valve/tr_valve_pipe.sql
lbartoletti: I don't understand why, but if it's not done BEFORE, the update is not done.

lbartoletti: OK Seems related to other trigger.

ordinary_data/valve/tr_valve_pipe.sql
lbartoletti: Without this test, you can let old count not modified.

lbartoletti:
Version 1.3.3 #250

PR #261, opened 2018-05-02
Customization

lbartoletti:
This PR adds the ability to use customization.

The test with travis is an example for that.

### DIFF
--- a/.build/travis_script.sh
+++ b/.build/travis_script.sh
@@ -76,4 +76,15 @@ printf "travis_fold:start:upgrade\nRun upgrade"
 pum upgrade -p qwat_prod -t qwat_sys.info -d $DELTA_DIRS
 echo "travis_fold:end:upgrade"
 
+# Extend qwat_prod with a customization
+printf "travis_fold:start:init-sire\nExtend database with a customization"
+$TRAVIS_BUILD_DIR/.build/customizations/sigip/init_sigip.sh -p qwat_prod -s 21781
+echo "travis_fold:end:init-sire"
+
+# Run upgrade with customizations/sigip/delta as an extra delta dir
+DELTA_DIRS="$DELTA_DIRS $TRAVIS_BUILD_DIR/.build/customizations/sigip/delta"
+printf "travis_fold:start:upgrade\nRun upgrade"
+pum upgrade -p qwat_prod -t qwat_sys.info -d $DELTA_DIRS
+echo "travis_fold:end:upgrade"
+
 exit 0
--- a/update/upgrade_db.sh
+++ b/update/upgrade_db.sh
@@ -7,14 +7,21 @@
 # Régis Haubourg
 # ##########
 
+GNUGETOPT="getopt"
+if [[ "$OSTYPE" =~ FreeBSD* ]] || [[ "$OSTYPE" =~ darwin* ]]; then
+	GNUGETOPT="/usr/local/bin/getopt"
+elif [[ "$OSTYPE" =~ openbsd* ]]; then
+	GNUGETOPT="gnugetopt"
+fi
+
 # Default values
 SRID=21781
 CLEAN=0
 LOCALDIRGIVEN=0
-LOCALDIR=/home/regis/OSLANDIA/projets_locaux/QWAT/local_update_dir_test/
 TMPFILEDUMP=/tmp/qwat_dump
 UPGRADE=0
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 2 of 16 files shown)


elemoine:
Thanks @lbartoletti. I'll have a look later today.

.build/travis_script.sh
```diff
@@ -76,4 +76,15 @@ printf "travis_fold:start:upgrade\nRun upgrade"
 pum upgrade -p qwat_prod -t qwat_sys.info -d $DELTA_DIRS
 echo "travis_fold:end:upgrade"
 
+# Extend qwat_prod with a customization
+printf "travis_fold:start:init-sire\nExtend database with ch_vd_sire"
```
elemoine: ch_vd_sire ?

.build/travis_script.sh
```diff
@@ -76,4 +76,15 @@ printf "travis_fold:start:upgrade\nRun upgrade"
...
+$TRAVIS_BUILD_DIR/customizations/sigip/init.sh -p qwat_prod -s 21781
```
elemoine: I'd name the file `init_sigip.sh` for consistency with `init_qwat.sh` and `init_sire.sh`.

customizations/sigip/damage/damage.sql
```diff
@@ -0,0 +1,32 @@
...
+  label_2_rotation double precision,
+  label_2_text character varying(120),
+  CONSTRAINT damage_pkey PRIMARY KEY (id)
+);
+
+GRANT SELECT, REFERENCES, TRIGGER ON TABLE qwat_od.damage TO qwat_viewer;
+GRANT ALL ON TABLE qwat_od.damage TO qwat_user;
+GRANT ALL ON TABLE qwat_od.damage TO qwat_manager;
+COMMENT ON TABLE qwat_od.damage
+IS 'pipe damage';
```
elemoine: Is this customization added only for the purpose of testing customization on Travis? I find it strange to add a `customizations` directory to the qwat repo while customizations are, by definition, outside the qwat repo.

update/upgrade_db.sh
```diff
@@ -7,14 +7,22 @@
...
+
 # Default values
 SRID=21781
 CLEAN=0
 LOCALDIRGIVEN=0
-LOCALDIR=/home/regis/OSLANDIA/projets_locaux/QWAT/local_update_dir_test/
 TMPFILEDUMP=/tmp/qwat_dump
 UPGRADE=0
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+#LOCALDIR="$( cd "$( "${SCRIPTDIR}/../customizations/sigip" )" && pwd)"
```
elemoine: Do you really want to keep that comment line?

update/upgrade_db.sh
```diff
@@ -28,25 +36,25 @@ case $key in
     -h|--help)
         echo "Arguments:"
         echo -e "\t-c|--clean\t\tCleans comp and test DB before starting"
-        # echo -e "\t-d|--localdirpath\tRuns a second update cycle with local customization delta files"
+        echo -e "\t-d|--localdirpath\tRuns a second update cycle with local customization delta files"
```
elemoine: In my opinion there should not be a second update cycle. The core, extensions and customizations should be updated as part of the same pum upgrade cycle.

customizations/sigip/damage/damage.sql
lbartoletti: Yes, it's only for Travis and set an example.

.build/travis_script.sh
lbartoletti: Right.

.build/travis_script.sh
lbartoletti: Oops, copy/paste

customizations/sigip/damage/damage.sql
elemoine: I'd suggest that we move the customizations directory to .build.

lbartoletti:
Thanks @elemoine

update/upgrade_db.sh
```diff
@@ -28,25 +35,25 @@ case $key in
     -h|--help)
         echo "Arguments:"
         echo -e "\t-c|--clean\t\tCleans comp and test DB before starting"
-        # echo -e "\t-d|--localdirpath\tRuns a second update cycle with local customization delta files"
+        echo -e "\t-d|--localdirpath\tAdds local customization delta files"
```
elemoine: What if I use both an extension and a customization? Will I be able to use upgrade_db.sh? I understand that upgrade_db.sh supports only one additional extension/customization, but I may be wrong.

lbartoletti:
release 1.3.3 #250

PR #266, opened 2018-08-16
add 'J' short_fr for Jauge in valve_function

lbartoletti:
qwat_vl.valve_function: add for vent valve the J value for the short_fr attribute, as it is the case for other

cc @kandre


lbartoletti:
release 1.3.3 #250

PR #268, opened 2018-08-16
add 25 as default pressure norminal for 'Fonte' in pipe_material

lbartoletti:
@kandre "Je propose de rajouter pour chaque entrée de fonte et de fonte-ductile) une nouvelle ligne de valeurs avec l'attribut pressure_nominal renseigné avec 25. L'utilisateur aurait alors le choix de choisir avec ou sans PN (= sous-entendu PN16). Selon les personnes du métier, les fontes en PN25 sont répandues chez les fournisseurs. Il est toujours possible d'obtenir des conduites et pièces (manchons, brides, etc) en fonte en PN25, donc je ne pense pas qu'on aura vraiment de couples de valeurs incohérents."

translation:
I propose to add for each input of cast iron and ductile iron a new line of values with the attribute pressure_nominal filled with 25, the user would then have the choice to choose with or without PN (= implied PN16). According to the people of the industry, PN25 fonts are widespread among suppliers. It is always possible to obtain pipes and parts (sleeves, flanges, etc) made of cast iron PN25, so I don't think we'll really have inconsistent pairs of values


tudorbarascu:
@lbartoletti Fine with me. However, I noticed that doing doing the update with the delta doesn't arrive to the same materials.

How I tested: created the database without your commit, ran the delta file (result is
problem - (Notice that PN25 is missing)

When doing the creation with your commit, the result is as it should.

I think the trigger tr_fancy_value doesn't function as good on updates as on inserts.

Shouldn't PUM have catched this?

lbartoletti:
Thanks @tudorbarascu 
You right _displayname_{en,fr,ro} are not updated since updating pressure_nominal doesn't trigger the function qwat_vl.pipe_material_displayname.

lbartoletti:
In discussions with the @haubourg, he told me that the PUM checks the structure of the tables, but not the content. This is an evolution to study, especially for our test case.

lbartoletti:
release 1.3.3 #250

PR #269, opened 2018-08-16
fix decimal value in pipe_material.sdr

lbartoletti:
Some decimal values (exported?) have a too digits after dot:
7.40000000000000036 for 7.40
13.5999999999999996 for 14.60


lbartoletti:
release 1.3.3 #250

PR #271, opened 2018-09-20
Update sire extension

kandre:
[WIP] This PR fixes model changes from older SIRE versions to SIRE 5.0.
https://www.vd.ch/fileadmin/user_upload/organisation/dse/scav/sire/Diff_mod%C3%A8le_4.8_5.0.xlsx

More fixes related to the code itself (and not the SIRE udpates) will come.

PR #272, opened 2018-09-24
Add 'IF NOT EXISTS' to qwat_od.damage

lbartoletti:
see https://github.com/qwat/qwat-data-model/pull/271#issuecomment-423189381


elemoine:
PR #273 suggests another way for fixing the "relation "damage" already exists" error, by properly structuring the sipip customization (in the same way as the ch_vd_sire extension).

lbartoletti:
target for release 1.3.3 #250

PR #273, opened 2018-09-24
Fix the sigip customization structure

elemoine:
This (hopefully) fixes the Travis build errors observed in #271, by properly structuring the sigip customization used in the tests.

Let's see what Travis says.

### DIFF
--- a/.build/customizations/sigip/init.sh
+++ b/.build/customizations/sigip/init.sh
@@ -73,4 +73,13 @@ if [[ "$DROPSCHEMA" -eq 1 ]]; then
          -c "DROP SCHEMA IF EXISTS qwat_sigip CASCADE"
 fi
 
-PGSERVICE=${PGSERVICE} SRID=${SRID} ${DIR}/inserts.sh
+# create the qwat_sigip schema
+psql service=${PGSERVICE} -c "CREATE SCHEMA IF NOT EXISTS qwat_sigip;"
+
+# add the qwat_od.damage table
+psql service=${PGSERVICE} -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/damage/damage.sql
+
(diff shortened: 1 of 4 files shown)


elemoine:
Ok, it works as expected. @haubourg @lbartoletti please merge if that makes sense to you.

There's one remaining error line 6182, but it's a different issue that should be addressed separately:

```
psql:/home/travis/build/qwat/qwat-data-model/roles.sql:3: ERROR:  role "qwat_viewer" cannot be dropped because some objects depend on it
```

.build/customizations/sigip/init.sh
```diff
@@ -73,4 +73,13 @@ if [[ "$DROPSCHEMA" -eq 1 ]]; then
          -c "DROP SCHEMA IF EXISTS qwat_sigip CASCADE"
 fi
 
-PGSERVICE=${PGSERVICE} SRID=${SRID} ${DIR}/inserts.sh
+# create the qwat_sigip schema
+psql service=${PGSERVICE} -c "CREATE SCHEMA IF NOT EXISTS qwat_sigip;"
+
+# add the qwat_od.damage table
+psql service=${PGSERVICE} -v ON_ERROR_STOP=1 -v SRID=$SRID -f ${DIR}/damage/damage.sql
```
lbartoletti: Maybe for the example (this directory is used for work but also as an example/demonstration of customizations) I will put the addition of tables in a script like 'insert_tables.sh'.
Here it doesn't make much sense because there is only one table, but it's just for example.
I'll let you judge for yourself.

elemoine: Yes, I agree. But for now I just wanted to structure sigip in the same way as ch_vd_sire.

lbartoletti:
target for release 1.3.3 #250

PR #274, opened 2018-09-24
Change upgrade_db.sh to avoid re-creating the roles

elemoine:
This is an attempt to eliminate the remaining build error:

```
psql:/home/travis/build/qwat/qwat-data-model/roles.sql:3: ERROR:  role "qwat_viewer" cannot be dropped because some objects depend on it
```


elemoine:
Say hi to the error-free build :smiley:

lbartoletti:
target for release 1.3.3 #250

Issue #178, opened 2017-07-03
Hydrant model inf sup


tudorbarascu:
Hello,

I was looking at hydrant model sup/inf values.
Are the inf ones supposed to be the underground models and the sup ones supposed to be for the 
"on the ground" ones?

I noticed there except the normal autre, inconnu and  a determiner values there are others common values like Hawle, Hawle_H4, Hinni, VonRoll.

Does it make sense to you to merge these into one table that has a boolean "underground" column to make things simpler?

tudorbarascu:
Comments in https://github.com/qwat/qwat-data-model/pull/179/commits/f19038e9d360107a22f12729482d38df0194c6f5

tudorbarascu:
Just noticed ``fk_model_sup`` and ``fk_model_inf`` are both set to not null in the hydrants table.
This means they stand for something else then what I imagined as a hydrant cannot be underground and on the ground at the same time, right?
Can someone shed a light on this?

3nids:
this is for above ground hydrants. They have two parts: the one under the ground, and the one above. Combinations of models is possible, hence the two fields.

tudorbarascu:
@3nids Thanks for the info.
So, as a summary it would seem we have the following cases:
* underground hydrant
* above the ground hydrant (single model)
* above the ground hydrant (combination of two parts/models)

tudorbarascu:
Taking another look at this, how is this supposed to model underground hydrants (which don't have two parts, at least here in Romania). Do you have underground hydrants made of two parts?
If not, then the data model should be enhanced as it currently enforces two parts also for the underground hydrants (with the NOT NULL constraints on both ``fk_model_sup`` and ``fk_model_inf`` columns).

I would like to propose that we drop the NOT NULL constraint for the ``fk_model_sup`` and add a CHECK constraint that still enforces the two part model for the case where the hydrant is above the ground.
This way we keep the current CHECKS except for the underground model which is single part.

Does it make sense to you @qwat/qwat-psc ?

tudorbarascu:
Something like this would suffice if we drop the ``fk_sup_model``:
```
ALTER table qwat_od.hydrant ADD CONSTRAINT above_ground_hydrant_has_2_parts_check check (
(fk_model_sup IS NOT NULL OR underground IS TRUE) -- If the hydrant is above ground fk_model_sup should be filled
);
```

[...]

PR #278, opened 2018-11-14
Enables underground hydrants to only have one part

tudorbarascu:
See discusion in #178


lbartoletti:
target for release 1.3.3 #250

PR #185, opened 2017-07-07
Improve survey points

tudorbarascu:
* added the fk_object_reference column
* created geometry index
* added a default sequence on id
* added en + ro translation of chevillère (I guess it's this.. right? https://en.wikipedia.org/wiki/Tape_measure#The_measuring_tapes)
* added Total Station as a type of survey points ( I guess you are also using Total Stations, right? )

I haven't created the delta file etc. yet as the automated system is not working currently and there are the others pull requests that may be merged or not. I will do it as soon as everything is clear.

Also, the ``COMMENT ON TABLE qwat_od.surveypoint IS 'Table for sampling points. Inherits from surveypoint.';`` is obsolete, right?

PR #279, opened 2018-11-14
Add default sequence to survey points

tudorbarascu:
Follow-up to #185 

It seems alright from my point of view. However, I didn't understood what was the purpose of the ``ALTER TABLE qwat_od.surveypoint ADD COLUMN id integer NOT NULL REFERENCES qwat_od.surveypoint(id) PRIMARY KEY;`` old construct, so there must be something I'm missing.
Thoughts?


lbartoletti:
target for release 1.3.3 #250

Issue #184, opened 2017-07-06
pipe protection SOMO value


tudorbarascu:
Hello,

Does anyone know what does ``SOMO`` stands for  (in qwat_vl.pipe_protection, SOMO PE, SOMO PVC, SOMO autre)?

ponceta:
SOMO tablet? Probably some switzery, I did not found any of these in the canplast catalog. @3nids you sure know better?

ponceta:
Ok, I was not so far from it.

SOMO stands for an old professionnal term (probably an old brand from canplast?) like actual gresintex http://www.canplast.ch/cataloguenew/pdf/ft_2.pdf

It's indeed a hard plastic protection pipe Pe for cables and PVC for water and wastewater.

ponceta:
Voici la réponse de Canplast à qui j'ai posé la question.

```
Bonjour,
 
SOMO est une ancienne marque disparue du groupe Bouygues qui ne caractérisait pas vraiment un type de tuyau, même si, dans la mémoire collective des années 80-90, cela voulait plutôt désigner, à tort, les tuyaux PE-câbles… 
 
A votre disposition pour plus de renseignements…
 
Meilleures salutations,
 
Solutions durables pour la gestion des eaux
Didier Lièvre
```

@tudorbarascu does this clarify your question?

tudorbarascu:
Yes!

PR #280, opened 2018-11-15
Describe what fk_protection SOMO value stands for

tudorbarascu:
Follow-up on #184


lbartoletti:
target for release 1.3.3 #250

Issue #139, opened 2016-12-12
Pully's EP_couverture


ponceta:
Hello,

A Pully on a une petite particularité, on a des points qui relèvent la profondeur des conduites. 

J'imagine que c'est une donnée utilisée principalemement pour mettre en évidence les endroits où les conduites sont enfouies plus profondément que d'habitude (ou sur des changements de matériau ou d'année). @qwat/board Est-ce qu'on est les seuls à avoir ça? Si oui @haubourg / @3nids, vous pensez que c'est mieux de les mettre où?

![image](https://cloud.githubusercontent.com/assets/8614223/21134414/1387f7d2-c11e-11e6-8e76-777ad82305e1.png)

A priori je pense rajouter soit rajouter un attribut "pully_depth" à la table qwat_od.survey_point ou créer une table de PointZ od.pully_depth si nous sommes les seuls à l'utiliser.

Sinon je suis ouvert à vos suggestions! :)

//

Hi,

In Pully we have points which are indicating pipe's depth (which is unusual or not). 

I'm thinking about where to put these since they add an additionnal information about pipe's depth at a precise node. Am I the only one who has such data?

If yes, I will probably add a "pully_depth" attribute to the qwat_od.survey_point table or create a pointZ table in qwat_od.pully_depth.

If not I'm open to suggestions.

tudorbarascu:
Hello,

Our workflow presumes that for each node on the pipe we also should survey the ground level point (the depth is derived from these two survey points).

I think that adding a depth attribute to the survey_point would be the best.
A ground level altitude column would also work fine as the depth could be presented as a virtual column.

haubourg:
Décision PSC 2017 03 07 : ajouter profondeur au surveypoint pour prochaine version de modèle

haubourg:
But surveypoints misses fk_object_referenced. Lausanne has a lot of other attributes for survey point

ponceta:
```
-- Table: qwat_od.surveypoint

-- DROP TABLE qwat_od.surveypoint;

CREATE TABLE qwat_od.surveypoint
(
  id serial NOT NULL,
  fk_survey_type integer NOT NULL,
  fk_worker integer,
  code character varying(50),
  description text,
  date date,
  fk_folder integer,
  altitude numeric(10,3) DEFAULT NULL::numeric,
  geometry geometry(PointZ,21781) NOT NULL,
  usr_session_name character varying(200),
  usr_no_pt character varying(50),
[...]

PR #283, opened 2018-12-06
Add depth information to surveypoints

ponceta:
Add depth information to surveypoints according to 

https://github.com/qwat/qwat-data-model/issues/139


lbartoletti:
target for release 1.3.3 #250

PR #286, opened 2019-01-09
Remove pum baseline on qwat_comp (upgrade)

### LINKED ISSUE #284: Remove pum baseline on qwat_comp (upgrade)
Baseline does not need to be set on qwat_comp when upgrading since it is already defined when initializing any qwat DB after v.1.3.0

pum baseline -p $PGSERVICE -t qwat_sys.info -d ${DIR}/update/delta -b $NUMVERSION

https://github.com/qwat/qwat-data-model/blob/master/init_qwat.sh#L265

We need to remove :

https://github.com/qwat/qwat-data-model/blob/master/update/upgrade_db.sh#L177-L183

lbartoletti:
Fix #284 
cc @ponceta


ponceta:
Thanks! :+1:

PR #291, opened 2019-03-04
Adds script to drop the customizations

lbartoletti:
Drops the schema with the views and drops the table qwat_od.dammage

cc @ponceta


ponceta:
Thanks @lbartoletti this has to be included to travis to have usable dumps?

Tested on my dev, works fine!

lbartoletti:
@ponceta not yet for the dumps, since we need to discuss about SIRE extensions.

ponceta:
Ok

Issue #294, opened 2019-03-08
Include FAE 600 to qwat_vl.pipe_material


ponceta:
FAE 600 is missing from qwat_vl.pipe_material

PR #295, opened 2019-03-08
Add FAE 600 to qwat_vl.pipe_material

ponceta:
Add new pipe material


lbartoletti:
fixes https://github.com/qwat/qwat-data-model/issues/294

PR #300, opened 2019-06-14
fix #299

### LINKED ISSUE #299: Schematic pipe views fails when parent pipes are merged
The `vw_pipe_schema` view relies on `vw_pipe_schema_merged` recursive view that merges on the fly all the pipes related via the `fk_parent` foreign key. 

In some cases, the pipes can be geographically connected or not, which leads either in generating a simple or a multi geometry.  
This throws then the following error in PG, which prevents it to be loaded from any client. The layer can't be loaded from QGIS at this point. 
I propose to release a bit the geometry type by forcing all geometries to be `Multi`

This line 
`  st_linemerge(st_union(pipe_find_parent.geometry))::geometry(LineStringZ,21781) AS geometry,`

would then  be switched to 

`st_multi(st_linemerge(st_union(pipe_find_parent.geometry)))::geometry(MultiLineStringZ,21781) AS geometry,`
 
Pull request is coming

haubourg:
I have a doubt about hardcoded SRID already working @lbartoletti . 
Let's see if Travis is happy


haubourg:
Repeat after me: 
"never put any view definition in a delta file"
"never put any view definition in a delta file"
"never put any view definition in a delta file"
"never put any view definition in a delta file"
"never put any view definition in a delta file"
"never put any view definition in a delta file"

PR #301, opened 2019-07-12
drop sire

lbartoletti:
Since we have dedicated repositories for extensions we have to cleanup qwat core by removing sire extension here.

cc @kandre @ponceta


ponceta:
This will be a 1.4.0 since we broke actual SIRE to new extension SIRE?

lbartoletti:
For me, yes

PR #302, opened 2019-07-21
be specific about altitude object reference regarding pipes

tudorbarascu:
fixes https://github.com/qwat/QWAT/issues/290


ponceta:
Thanks @tudorbarascu

PR #303, opened 2019-07-24
Archive metaproject submodule

3nids:
Since there is only one script used in metaproject repository and by QWAT only, we think it's better to have it placed here and to archive the repository.

discussed with @ponceta

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "metaproject"]
-	path = metaproject
-	url = https://github.com/opengisch/metaproject.git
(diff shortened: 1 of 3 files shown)


3nids:
Note: regarding the future of these highly complex and unmaintainable scripts, I would recommend to used the refactored version of the other script (the non multiple inheritance one) and writing SQL with the basic Python tools from pirogue.
Here is an example in QGEP for a complex view: https://github.com/QGEP/datamodel/blob/master/view/vw_qgep_reach.py
The chosen approach rely on simpler tested methods and use pseudo-SQL syntax which is much easier to maintain/adapt.

lbartoletti:
Good for me. Thanks @3nids .

Yes I agree, We must homogenize QWAT and QGEP and use the same tools.

PR #310, opened 2019-10-24
Project planned

tudorbarascu:
Up until now we had only one "project" status so there wasn't any way
to distinguish a planned project from an project in execution

Important:

The "project" value has now the "planned project" meaning and there's
a new value for status (1308) that represents the project in execution.

Please check for the wording in french..

In the qgep model there are also these values, which we don't have in the qwat model.
Should we include them also?

```
  | operational.tentative | en_service.provisoire
  | operational.will_be_suspended | en_service.sera_supprime
  | abanndoned.suspended_not_filled | abandonne.supprime_non_demoli
  | abanndoned.suspended_unknown | abandonne.supprime_inconnu
  | abanndoned.filled | abandonne.demoli
```


tudorbarascu:
gentle ping @kandre @ponceta @haubourg @lbartoletti

update/delta/delta_1.3.6_add_status_project_execution.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO qwat_vl.status (id, active, functional, value_en, value_fr, value_ro, description_en, description_fr, description_ro) VALUES (1308, false, false, 'project in execution',        'projet en execution',       'Proiect în execuție',      '', 'L''objet est un projet validé en cours de réalisation', 'Obiectul face parte dintr-un proiect care este în curs de execuție');
```
lbartoletti: Can you uppercase for FR and EN. And add long description for EN ?

kandre:
Nice idea to include that in the model. At SIGE, we use a custom value "waiting" instead for 1308, which is pretty much the same. French translation looks good to me.

update/delta/delta_1.3.6_add_status_project_execution.sql
tudorbarascu: Sure about the en description. For the uppercase part.. I've uppercased for all the values (not only 1308).

Commens addressed.

lbartoletti:
Thank you @tudorbarascu

PR #325, opened 2020-04-21
Add a generic pipe for unknown PE type (id=9999)

nicolas-sitylb:
Notice: I don't speak RO so feel free to fix the translation (I wrote the extra suffix in `value_ro` in english).


value_lists/vl_pipe_material.sql
```diff
@@ -561,6 +562,7 @@ INSERT INTO qwat_vl.pipe_material (id, vl_active, short_en, value_en, short_fr,
...
 INSERT INTO qwat_vl.pipe_material (id, vl_active, short_en, value_en, short_fr, value_fr, short_ro, value_ro, diameter, diameter_nominal, diameter_internal, diameter_external, pressure_nominal, sdr, wall_thickness) VALUES (7629, true, '', '', 'PE80', 'Polyéthylène 80', 'PE80', 'Polietilenă 80', '90', NULL, 81.400, 90.000,  6, 21, 4.3);
 INSERT INTO qwat_vl.pipe_material (id, vl_active, short_en, value_en, short_fr, value_fr, short_ro, value_ro, diameter, diameter_nominal, diameter_internal, diameter_external, pressure_nominal, sdr, wall_thickness) VALUES (7630, true, '', '', 'PE80', 'Polyéthylène 80', 'PE80', 'Polietilenă 80', '110', NULL, 99.400, 110.000,  6, 21, 5.3);
+INSERT INTO qwat_vl.pipe_material (id, vl_active, short_en, value_en, short_fr, value_fr, short_ro, value_ro, diameter, diameter_nominal, diameter_internal, diameter_external, pressure_nominal, sdr, wall_thickness) VALUES (9999, true, '', '', 'PE inconnu', 'Polyéthylène de type inconnu', 'PE unknown', 'Polietilenă unknown type', NULL, NULL, NULL, NULL, NULL, NULL, NULL);
```
haubourg, line 565: Hi - by consistency of serial id, maybe 7699 would be a bette id value? Polyéthylène is all in the 7xxx range

nicolas-sitylb: Do you have a description of those different ranges, I can guess most of them, but I do not like "guessing" stuff. ;-)

haubourg: nope, I don't know of any specification for id ranges.

sjib: @3nids Do you know how the id's were set at the beginning?

tudorbarascu: @nicolas-sitylb For PE we already have id 5501 who takes care of the unknown value. I think your PR should update those values to "unknown etc."
You can replace the Ro value from ``'Polietilenă unknown type'`` with ``PE necunoscută``

Thank you!

ponceta: What I see here is that we have two PE with unknown diameters : 

- 1st is id 5501 for PE (100) with unknown diameter.
- 2nd is id 5550 for PE 80 with unknown diameter.

I would not be in favor of adding an unknown PE material, since "plastic" generic material type is a synonym of PE unknown. In that case it would be better to use either 

 - 3rd id 5651 plastic with unknown diameter
or
- 1st id 5501 for PE (100) with unknown diameter as default value when PE type is unknown.

ponceta, line 565: Therefore:
```suggestion
```

tudorbarascu: Hi, I always took the plastic as also containing PVC or GRP. If we differentiate between PE(100) and PE80 then we should also differentiate the unknown values.

nicolas-sitylb:
@tudorbarascu I updated the Romanian translation according to your proposal.
I also re-synchronized my `master` branch according to the current state of the `qwat:master`.

Issue #340, opened 2021-09-13
Update script fails because dir is not mentionned in upgrade script


ponceta:
- [ ] Fix update/upgrade_db.sh to consider delta dir

pum test-and-upgrade: error: argument -d/--dir: expected at least one argument

![image](https://user-images.githubusercontent.com/8614223/133049233-078d6f53-b435-4b68-890d-658662e79207.png)

ponceta:
https://github.com/qwat/qwat-data-model/blob/master/update/upgrade_db.sh#L25

Could be set by default as deltas/ or update/deltas depending on our standard update procedure and should be overwritten when specified though an env variable from the user.

PR #342, opened 2021-09-13
Test default directory for delta folder

ponceta:
To check github actions


lbartoletti (approved):

ponceta:
Should fix https://github.com/qwat/qwat-data-model/issues/340

According to doc :

https://qwat.github.io/docs/master/fr/html/developer-guide/upgrade.html#migrate-an-existing-database

ponceta:
@olivierdalang could you check this small fix and merge it if all green?

PR #343, opened 2021-10-02
Fix reverse pipe should switch fk_node_a and fk_node_b, currently it does not

tudorbarascu:


ponceta:
@tudorbarascu could you please rename your delta to 1.4.0_002_fix_pipe_reverse_nodes.sql so we can merge this for next release?

ponceta:
@tudorbarascu Thank you!

PR #349, opened 2021-12-21
Fix orientation of element part when updating pipe or element part

benoitblanc:
## Description

When user wants to connect an element part to a pipe, orientation of the element is wrong, fixing angle to 0 or some wrong value.

### Before
qwat

### After

demo_qwat_fix_orientation

To do so, I fixed the node taken into account when updating pipe in trigger (NEW instead of OLD).

Then, for the element part update, I need to call a function in auto-generated trigger `qwat_od.ft_element_part_update`. So I have created a `custom_update` in `ordinary_data/views/inheritance/od_all_inheritance.py` and I wanted to handle it in `ordinary_data/views/inheritance/pg_inheritance_view_recursive.py` in `sql_join_update_trigger` method (as there is the same thing with `custom_delete`). But I have an error when I try to generate views :

```
/QWAT/data-model/ordinary_data/views/insert_views.sh: ligne 15: /usr/bin/psql: Liste d'arguments trop longue
```

So I am wondering how I can add an argument, something is missing but I can't find what... @3nids I need some help to understand how to add custom code in auto-generated triggers.

Thanks


ordinary_data/functions/fn_element_part_set_orientation.sql
```diff
@@ -0,0 +1,108 @@
...
+                bool_or(coalesce(schema_force_visible,pipe_function.schema_visible)) AS schema_visible
+            INTO _grouped
+            FROM qwat_od.pipe
+                INNER JOIN qwat_vl.status ON pipe.fk_status = status.id
+                INNER JOIN qwat_vl.pipe_function ON pipe.fk_function = pipe_function.id
+            WHERE (pipe.id = (SELECT fk_pipe FROM qwat_od.vw_element_part WHERE id = _element_id))
+                AND status.active IS TRUE;
+
+               
+        raise notice 'Pipes associés à %, %', _element_id, _grouped;
```
3nids, line 34: this should be in English

ordinary_data/functions/fn_element_part_set_orientation.sql
```diff
@@ -0,0 +1,108 @@
+CREATE OR REPLACE FUNCTION qwat_od.fn_element_part_set_orientation(_element_id integer)
```
3nids, line 1: this looks similar to https://github.com/qwat/qwat-data-model/blob/master/ordinary_data/functions/fn_valve_set_orientation.sql
would there be a chance to avoid code duplication?

ordinary_data/pipe/od_pipe_geom.sql
```diff
@@ -90,10 +90,10 @@ CREATE OR REPLACE FUNCTION qwat_od.ft_pipe_node_type() RETURNS TRIGGER AS
 		END IF;
 		IF TG_OP = 'UPDATE' THEN
 			IF NEW.fk_node_a <> OLD.fk_node_a THEN
-				node_ids := array_append(node_ids, OLD.fk_node_a);
+				node_ids := array_append(node_ids, NEW.fk_node_a);
```
3nids, line 93: good catch

3nids (commented):
I am not sure for the custom function. Sylvain originally wrote the Python code, and I am not really familiar with the code.

Other thing, Sylvain also wrote some tests for this in the past: https://github.com/qwat/qwat-data-model/blob/37ca3c6f6f8403371d60244d84246e0c9931b5d5/tests/test_node_orientation.sql It would be nice to get them running on the CI + add yours.

ordinary_data/functions/fn_element_part_set_orientation.sql
benoitblanc: Yes it is quite similar, but as it is not the same object there are some differences. Maybe we could avoid code duplication but it needs more thinking about it (maybe consider other objects that need to be oriented) ?

benoitblanc:
> I am not sure for the custom function. Sylvain originally wrote the Python code, and I am not really familiar with the code.

It was the right way to add custom code in generated trigger, but I add two `;;` generated. One in `od_all_inheritance.py` and one in `fn_element_part_set_orientation.sql` :facepalm:

ponceta:
Since CI is functionnal again, I would try to get this fixed if possible.

@kandre can you follow this?

benoitblanc:
@ponceta I have rebased this branch on master, added deltas and added one test to update orientation by moving a node of a pipe.

ponceta:
Thank you @benoitblanc this is great!

Issue #350, opened 2022-02-10
The view vw_export_pipe is always in crs 21781


tproduit:
The scripts which generates this view doesn't take into account the crs set by the user. It always set the crs as 21781.

ponceta:
![Screenshot_20220216-130015474.jpg](https://user-images.githubusercontent.com/8614223/154260472-3ff6cf67-1f1d-4d99-b6ae-91f5b332f4be.jpg)

This is indeed hard coded and should be using the SRID stored either in the database or SRID supplied as parameter !

PR #351, opened 2022-03-03
Export pipe srid

tproduit:
To solve issue #350


ponceta (approved):
Thank you @tproduit for this fix.

Issue #348, opened 2021-12-14
Continuous Integration CI : Unit test are no longer running on pull requests


ponceta:
Coherency control and sql statements are not tested anymore in an automatic way which can result in problems when upgrading to latest versions.

ponceta:
@olivierdalang thank you for what has been achieved!

We still have :

- [ ] To look why the specified postgresql is not used in the tests
- [ ] To have similar docker images tested like for qgep
- [ ] To have similar behaviour to be able to start and redo the testing on a PR.

![image](https://user-images.githubusercontent.com/8614223/201276441-58d7e189-78fa-4884-8815-91d14056363d.png)

ponceta:
- [ ] Run upgrade tests can have errors and still be flaged as successful!
- [ ] roles have to be created before restoring the database

![image](https://user-images.githubusercontent.com/8614223/201289948-934f03e1-8729-4c2f-8fc9-f7fb6356d608.png)

https://github.com/qwat/qwat-data-model/actions/runs/3443193005/jobs/5744493765#step:9:2219

olivierdalang:
> * [ ]  To look why the specified postgresql is not used in the tests

I think it runs against 9.6, but with a 12.12 client (installed in the `apt-get install postgresql-client` step). IMO this is fine, as it will still correctly test all database side logic, and newer clients do support older backends.

> * [ ]  To have similar docker images tested like for qgep

Yes that would be nice (also worth a rewamp in the QGEP side), but IMO not really related to testing ? I see it more a nice to have for quick deployment of a demo database.

> * [ ]  To have similar behaviour to be able to start and redo the testing on a PR.

Not sure what you mean here ?

> * [ ]  Run upgrade tests can have errors and still be flaged as successful!

Are you talking about the errors in restore (which could be harmless, as the role error you mentionned), or in the actual pum upgrade ? (the later should definitely be catched, if it's not the case can you point me towards where you saw this happen ?).

For now the results of the upgrade are just tested with pum check, but that does not take any content into account. Not sure if the QWAT test suite relies on the demo data, if yes we could also run tests agains the upgraded db to try catching some data related upgrade error.

> * [ ]  roles have to be created before restoring the database

I think it would mean add the `--create-roles` to init_qwat.sh. But it seems none of the test actually make any usage of these roles. Thus I think we can just leave it like that as long as roles are considered an optional component of QWAT (btw I think it would be a good move to move them to core and to start to test them as well, same applies to QGEP).

[...]

PR #354, opened 2022-04-27
Windows support

### LINKED ISSUE #353: `delta_1.4.0_003_create_fk_leak_type` fails with `column "fk_leak_type" does not exist`
Indeed, just realized also that master does not work here (`delta_1.4.0_003_create_fk_leak_type` fails with `column "fk_leak_type" referenced in foreign key constraint does not exist`). 

This would have been catched by CI easily, but it seems tests don't run (see https://github.com/qwat/qwat-data-model/issues/348)

olivierdalang:
This translates shell scripts to python, so migrations can be run on windows too.

I tested it locally. Some notes/comments/questions:

- On the test DB, sigip views must be dropped manually (otherwise pre-all fails with `cannot drop view qwat_od.vw_export_pipe because other objects depend on it`). What is the status of these views ? Is it normal that pre-all does not drop them and post-all recreate them ?

- Before merging this, we should make CI work again, as otherwise it's too risky to do this type of work (see https://github.com/qwat/qwat-data-model/issues/348)

- Also, before this is usable, another blocker must be fixed: https://github.com/qwat/qwat-data-model/issues/353


3nids:
> On the test DB, sigip views must be dropped manually (otherwise pre-all fails with cannot drop view qwat_od.vw_export_pipe because other objects depend on it). What is the status of these views ? Is it normal that pre-all does not drop them and post-all recreate them ?

I think we should not have views in extra schemas in the test DB. They might need to be removed?

> Before merging this, we should make CI work again, as otherwise it's too risky to do this type of work

Indeed. @ponceta is there any plan here?

> Also, before this is usable, another blocker must be fixed: https://github.com/qwat/qwat-data-model/issues/353

@ponceta would you look to this one too?

ponceta:
To make things right I would propose as guidelines from experience:

- [ ] export views must ALL be in an export schema (**qwat_export**) (like QGEP)
- [ ] custom views must not reference export views (avoid locks as much as possible)
- [ ] documentation and process should take into account a few custom views as examples -> rewrite some (not all) of the sigip views into custom views

PR #355, opened 2022-05-05
Fix delta constraint around leak types

ponceta:
Fix delta constraint around leak types


olivierdalang (approved):
Ok for me. I guess this fixes #353 ?

You marked it as WIP though, is it ready to merge ? (if so feel free to merge it)

PR #359, opened 2022-06-24
Inheritance of attributes on nodes

benoitblanc:
This PR aims to set some attributes to nodes regarding connected pipes.

This concerns only `node` table.

There are several attributes to inherit from connected pipes :
- [x] `fk_status` : fk_status integer; -> values (qwat_vl.status)
  The inherited attribute on the node should retrieve the fk_status value of the connected pipe. In case several pipes are connected, the retrieved value must respect a priority order.
Each user-administrator should be able to define this priority order for each status in the database, including for custom values. Indeed, status values are not used in the same way by different organisations.
Finally, a priority order should be defined by default in the data model and would be as follows

  * En service : prio 1 (in use)
  * En attente : prio 2 (project in execution ?)
  * Projet : prio 3 (project planned ?)
  * Hors service : prio 4 (out of service)
  * Désaffecté : prio 5 (shut down)
  * Abandonné : prio 6 (disused)
  * Détruit : prio 7 (destroyed)
  * Fictif : prio 8 (fictional)
  * Autre : prio 9 (other)
  * Inconnu : prio 10 (unknown)
  * A déterminer : prio 11 (to be determined)

- [x] `fk_distributor` : fk_distributor integer[]; -> values (qwat_od.distributor)
    The inherited attribute on the node should take all the values of the distributor field of the connected pipes as a list
- [x] `fk_pressurezone` : fk_pressurezone integer[]; -> values (qwat_od.pressurezone)
  Note that this attribute is already inherited in the node table (qwat_od.node) as an integer.
Currently, the pressure zone attribute is obviously currently inherited on the node by geometric intersection with the pressure zones. When a node is at the intersection of one or more pressure zones, we observe that the retrieved value is random (the largest id, it would seem).
The inherited attribute on the node should take all the id values of the intersecting pressure zones as a list


ponceta:
Looks almost fine, I'll be happy to test it soon. 

@benoitblanc could you have a look at the CI tests that are now failing because of our adaptations.

https://github.com/qwat/qwat-data-model/tree/master/tests

benoitblanc:
Hi @ponceta sorry for the delay, I have missed notification.

I pushed new commits to have tests succeeded. I think it is ready to test :smile:

ponceta:
Delta files have to be renamed to 1.4.0_XXX since we won't push a 1.3.7 version.

ponceta:
Feels good on QGIS. Once we push this it would be nice to update the qgs project with these modifications.

Here a sample of how it looks in a form : Multiple distributors connected on a single Node, Status prioritized.

image

PR #363, opened 2022-07-11
Avoid references to productions views directly (prefer tables)

ponceta:
Example of how export views should look like VS reference to productions views like vw_element_hydrant


ponceta:
@olivierdalang this should avoid any troubles with having locks on editing views.

IMHO it would be to extend this to any export views as said in

https://github.com/qwat/qwat-data-model/pull/354

ponceta:
No delta is required (views only)

olivierdalang:
A bit difficult to review for me, as I have 0 context as to how this is used. Purely from the code it looks fine though, so I'd say if you tested and it seems to still work, it's ok to merge.

Merging this should be coordinated with https://github.com/qwat/qwat-data-model/pull/365 (to ensure we remove the hotfix for sigip, which otherwise will be untested)

PR #364, opened 2022-08-04
[CI] (re?)enable tests in CI

olivierdalang:
Seems to pass.

Had to do some testing infrastructure related changes, and two actual fixes:
- specify a precision for test data inserted in qwat_od.cover (otherwise it fails because of non null - I guess it's a relatively new constraint ?)
- add `vl_leak_type.sql` in `init_qwat.sh` (seems that was forgotten in recent changes)

I'm not familiar with the test suite however, so not exactly sure how much we can rely on them.

What seems **not tested at all** is the upgrade scripts. The test use `init_qwat.sh` which runs the SQLs directly. We should probably also include a test that starts with version 1.0, and upgrades with pum, and then checks the consistency of the upgraded model with the `init_qwat` model, like we do for QGEP. This is done a followup PR since this is likely to require actual fixes to the datamodel : https://github.com/qwat/qwat-data-model/pull/365


ponceta:
@olivierdalang Looks good to me, I've asked a review from oslandia if that's ok for you.

olivierdalang:
👍

tests/tests.sh
```diff
@@ -43,7 +42,7 @@ EXITCODE=0
 for f in ${TESTS}; do
     echo -n "Running $f ... "
     fo="/tmp/${f}.txt"
-    psql service=${PGSERVICE} ${PGOPTIONS} -tA -f ${DIR}/$f >$fo 2>&1
+    psql service=${PGSERVICE} -c "SET lc_messages TO 'en_US.UTF-8'" -c "SET client_min_messages TO ERROR" -tA -f ${DIR}/$f >$fo 2>&1
```
lbartoletti, line 45: you could use `C.UTF-8`

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,43 @@
+name: Run tests
+on: push
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgis/postgis:9.6-2.5
```
lbartoletti, line 10: If possible, later, we should tests with a matrix.

However, Postgresql10 (9.6 is not supported by PostgreSQL) and PostGIS 3.0 should be mandatory now.

lbartoletti (approved):
LGTM, only two minor comments

.github/workflows/run_tests.yml
ponceta: Since april 2021 we have some Execute statement in qwat-datamodel code what implies that 9.6 is now fully unsupported on newer versions of QWAT min postgres version is now 10, no special requirements on postgis side to my knowledge.

https://github.com/qwat/qwat-data-model/search?q=execute

https://www.postgresql.org/docs/current/sql-syntax-calling-funcs.html

ponceta: If the CI works correctly, this should fail :)

ponceta:
@olivierdalang Let's get merging?

.github/workflows/run_tests.yml
olivierdalang: @ponceta Are you sure it should fail ? `EXECUTE` existed on 9.6, aren't you talking about `EXECUTE FUNCTION` (which I think only comes in pg11, and is addressed in followup pr see https://github.com/qwat/qwat-data-model/pull/365/commits/ed4fcd3f2e3d1e66da543fe78074cadf5523385d)

However I agree we should bump to more up to date postgres, but would suggest to do that separately (so that this really is about reenable tests we used to run on travis as is)

tests/tests.sh
olivierdalang: done

ponceta:
thank you @olivierdalang

PR #365, opened 2022-08-04
[CI] include pum check in CI to test datamodel upgrades

olivierdalang:
**Dependencies**
- [x] This is based on https://github.com/qwat/qwat-data-model/pull/364 which should be reviewed/merged first.
- [x] This requires a new release of pum with following fixes:
  - [x] https://github.com/opengisch/pum/pull/102 (otherwise there are false positives)
  - [x] ⚠️⚠️⚠️ **CRITICAL - DO NOT MERGE WITHOUT THIS**⚠️⚠️⚠️ https://github.com/opengisch/pum/pull/81 (otherwise changes in existing deltas will be reapplied and fail/mess things up)

**TODO**
- [x] Fix more inconsistencies highlighted by latest version (>0.9.14) of pum

**Notes**

I think the tests do work as expected, and highlighted the following issues (which are fixed in this PR):

- the qwat_sigip schema is present in the releases, but does not seem supported by the deltas. For now, I'm just manually dropping the schema in the CI before running pum check. But this schema should be removed from the dumps, or the deltas fixed to work with it.
- tables related to documents were missing in the initialization scripts
- pre/pos-all scripts could silently fail (because of `os.system` that does not raise exceptions on failure)
- a minor inconsistency in the name of an index
- a script that was not compatible with postgres<=10 (which should still be supported)
- this does not yet include anything about making release (which still uses the same travis script that at also tests upgrades with pum check). Removing this overlap and fixing the release (e.g. remove sigip extension) will be done in a separate PR.


ponceta:
https://github.com/qwat/qwat-data-model/pull/363 Should fix qwat_sigip related issues. If not I will be glad to contribute to fix it properly to ensure small customizations are supported.

ponceta:
@olivierdalang review asked to oslandia if that's ok for you?

tests/tests_upgrade.sh
```diff
@@ -0,0 +1,72 @@
...
+
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+INIT_DB=1
+
+while [[ $# > 0 ]]; do
+key="$1"
+case $key in
+    -h|--help)
```
lbartoletti, line 15: bash builtin long option names with the double-dash prefix. This is an extension from GNU getopt (on macos and bsd-like you have to install gnu-getopt).

More details: https://github.com/qwat/qwat-data-model/pull/228
and from Denis https://stackoverflow.com/a/37485578 :)

lbartoletti (approved):
LGTM, with previous comments and one for a gnu-ism

tests/tests_upgrade.sh
olivierdalang: I copied this from `tests.sh`, thus I suggest to keep it like that for consistency

(to be honest, no idea what this is about :-D)

olivierdalang:
I think this can be merged provided we're aware of this risk:

I'm not 100% sure this wont cause havoc for those who are not using the latest pum version, since there is a change to migrations files.

For the change in `update/delta/delta_1.2.8_schema_visible_perf.pre.py`, I guess pum doesn't keep track of its checksum anyway (it's a `pre` migration), so I **think** it would not cause issues. If we revert that, we may have silently failing issues.

For the changes in 1.4.0_*, since not release yet, I guess we can consider they are in no one's production database. If we revert that, we loose postgres 9.6 compatibility.

ponceta:
Thank you @olivierdalang ! We will fix the remaining bug and stuff when they appear.

PR #369, opened 2022-11-18
Drop view permissions to avoid logging

ponceta:
Until rights management is fixed in CI


ponceta:
Avoid useless logs in CI

ponceta:
(Does not apply to production views since these have inherited rights from schema definition)

PR #345, opened 2021-10-20
Network intervention simulation

benoitblanc:
This pull request aims to add some features for the user to simulate network incidents :
* network leakage
* network interruption
* network pollution

PR #375, opened 2023-01-19
Update scalability tests with new field type (fk_distributor)

benoitblanc:
Fix an error when running scalability tests because new field type for `fk_distributor` has not been updated in PR https://github.com/qwat/qwat-data-model/pull/359 as it was the case in commit https://github.com/qwat/qwat-data-model/pull/359/commits/904b154b64d75eba5ac69af1c28b73310af0632d for other tests.

See comment https://github.com/qwat/qwat-data-model/pull/345#issuecomment-1386995958


ponceta (approved):
LGTM

Issue #107, opened 2016-03-01
Proposal/Enhancement for Nominal pressure


tudorbarascu:
The materials and the nominal pressure combinations do not fully cover whatever is in the field.
I remembered that we even considered dropping the PN column.

My proposal is to begin using the `pressure_nominal` column with the following convention:
- for a PE 110 Pipe of 10 PN material the `pressure_nominal` column shouldn't be used (NULL) as we already have it from the materials table. However, in case the pipe is old or we consider it's not in the best shape due to various reasons and we want to restrict the nominal pressure to `8` let's say, by filling the `pressure_nominal` column we **override by convention** the PN.
So, although the pipe material is `10`, we consider the pipe to have PN of `8`.
- This would also solve the cases where the material tables is lacking some combinations between materials and PN (for various reasons: no standards etc.) where we can put the PN in the pressure_nominal.

Related to #87 and #21 

Thoughts anyone?

rbovard:
> Also, IMHO we shouldn't enforce the pressure_nominal default to 16 (or any other) in the db model but this should be moved to the QGIS side of the project.

I totally agree, it's dangerous to keep this behaviour since this attribute is not linked to the nominal pressure chosen in material!

rbovard:
I've opened a PR to remove the default value https://github.com/qwat/qwat-data-model/pull/391

PR #391, opened 2024-12-06
Do not use an arbitrary default value for PN

rbovard:
Related to https://github.com/qwat/qwat-data-model/issues/107#issuecomment-437679392

Need some hint on the failing test.


sjib:
You need to create a delta file for this change.

ponceta:
I pushed the corresponding delta.

PR #393, opened 2024-12-20
Put handle_geometry column after main geometry column

### LINKED ISSUE #380: [ENH] Put the `geometry` field of the qwat_od.valve table BEFORE all other geometry fields.
Hello,

What is the purpose of the `handle_geometry` field in the `qwat_od.valve` table: https://github.com/qwat/qwat-data-model/blob/d552ec9567a059155d2c1bd78c7425eacc7efbf3/ordinary_data/valve/od_valve.sql#L23 ?

If the `geometry` field in the `qwat_od.valve` table is intended to store the real geometry of the features and if the importance of the `handle_geometry` field is less than the `geometry` field in this table, I would suggest to place the `geometry` field _BEFORE_ all other geometry fields.

Why?

Because some software is unfortunately not able to deal with other than _the very first_ geometry column encountered in a PostGIS table. Therefore, as that _very first_ geometry column in the `qwat_od.valve` table is currently `handle_geometry` and because it occurs to be empty in most data sets which store the real geometry of features in the `geometry` column (and because such other software doesn't offer a solution to manually catch the real geometry), I would suggest to put the `geometry` field _BEFORE_ all other geometry fields in the `qwat_od.valve` table.

Thanks.

Warm Regards.

rbovard:
Fix #380

### DIFF
--- a/ordinary_data/valve/od_valve.sql
+++ b/ordinary_data/valve/od_valve.sql
@@ -20,7 +20,6 @@ ALTER TABLE qwat_od.valve ADD COLUMN fk_nominal_diameter    integer;
 ALTER TABLE qwat_od.valve ADD COLUMN closed                  boolean default false;
 ALTER TABLE qwat_od.valve ADD COLUMN networkseparation       boolean default false;
 ALTER TABLE qwat_od.valve ADD COLUMN handle_altitude         decimal(10,3);
-ALTER TABLE qwat_od.valve ADD COLUMN handle_geometry         geometry(PointZ,:SRID);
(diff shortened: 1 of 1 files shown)


ponceta:
This could only work for `new` installations and will differ from `existing` installations.

The only we can handle that in a delta is to copy the column into a new column, drop the actual column, rename the new column with the old name and if necessary alter business logic linked to the old column if necessary. And it will be placed after any eventual.

PUM does not check the column order that's why the upper CI test is green.

https://stackoverflow.com/questions/285733/how-do-i-alter-the-position-of-a-column-in-a-postgresql-database-table

PR #394, opened 2025-02-03
Reorganize datamodel scripts

ponceta:
Rework the datamodel scripts structure to ease the db creation and CI


ponceta:
Looks like something changed here :

image

https://github.com/qwat/qwat-data-model/blob/master/update/delta/delta_1.4.0_013_fix_orientation.sql

ponceta:
@kandre tu fais tourner qwat sur quelle version de postgres / postgis en ce moment?

3nids (approved):
I had a quick look and it looked good to me. In any case, I don't see any risk here regarding the datamodel.

PR #407, opened 2025-09-12
Update od_all_inheritance.py

tproduit:
Allows the user to set manually fk_pipe for parts. I'm not sure of the syntax to be used in this file.

### DIFF
--- a/datamodel/ordinary_data/views/inheritance/od_all_inheritance.py
+++ b/datamodel/ordinary_data/views/inheritance/od_all_inheritance.py
@@ -127,7 +127,7 @@ children:
             pkey: id
             c_alter:
                 fk_pipe:
-                    write: qwat_od.fn_pipe_get_id(NEW.geometry)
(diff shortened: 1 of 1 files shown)


datamodel/ordinary_data/views/inheritance/od_all_inheritance.py
```diff
@@ -127,7 +127,7 @@
             pkey: id
             c_alter:
                 fk_pipe:
-                    write: qwat_od.fn_pipe_get_id(NEW.geometry)
+                    write: CASE WHEN NEW.fk_pipe IS NULL THEN qwat_od.fn_pipe_get_id(NEW.geometry) ELSE NEW.fk_pipe END
```
ponceta: Looking at Line 133, you have probably to add ""

ponceta:
```suggestion
                    write: "CASE WHEN NEW.fk_pipe IS NULL THEN qwat_od.fn_pipe_get_id(NEW.geometry) ELSE NEW.fk_pipe END"
```

PR #427, opened 2026-05-06
Upgrade to latest ubuntu 26.04

ponceta:
Upgrade to latest ubuntu 26.04


ponceta:
It looks like github actions is not yet 26.04 ready.

PR #433, opened 2026-08-13
New version of release workflow

ponceta:
New version of release workflow


sjib:
@ponceta Still valid?

ponceta:
This will be in TMMT release

